PR #1, opened 2020-07-13
Add classes to 1) represent a macro's properties and 2) initialize firebase and do a query for a macro.

SydneyCaulfeild:


SydneyCaulfeild:
I have renamed some directories. This is what the folders now look like: 
image

In the datastorage folder, I have added an interface called DataStorage with one method named getMacro that should receive a macro's creator and name and return the macro. FirebaseDataStorage is a class that implements DataStorage using firebase.

I have also added a macro_classes folder which contains three classes that together will represent a macro.

src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
```diff
@@ -0,0 +1,72 @@
...
+    public FirebaseDataStorage () {
+        try {
+            this.db = initializeFirebase();
+        }
+        catch (FileNotFoundException e) {
+            System.out.println("Service key not found.");
+        }
+        catch (Exception e) {
+            System.out.println("Firebase not initialized.");
+        }
```
SydneyCaulfeild: Is there a better way for me to handle this exception? For instance, should I be doing System.exit instead of println()?

src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+ 
+    private Firestore initializeFirebase() throws Exception {
+        FileInputStream serviceAccount;     
+        serviceAccount = new FileInputStream("key.json");
```
SydneyCaulfeild, line 67: Do I need to close this input stream?

src/main/java/com/google/flourbot/macro_classes/Macro.java
```diff
@@ -0,0 +1,47 @@
...
+    public String getDocId () {
+        return docId;
+    }
+
+    public void createMacroAction (String[] columnValue, String sheetAction, String sheetUrl, String actionType) {
+        macroAction = new Action(columnValue, sheetAction, sheetUrl, actionType);
+    }
+
+    public void createMacroTrigger (String triggerCommand, String triggerType) {
+        macroTrigger = new Trigger (triggerCommand, triggerType);
```
SydneyCaulfeild: Is it ok for the trigger and action properties of a macro to be defined in their own methods? Or should the trigger and action for a macro could be defined in the constructor for the macro? The constructor would need to accept more arguments.

src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
tsa87: My opinion is these exceptions could be propagated with `FirebaseDataStorage () throws ...` so the servlet that uses `FirebaseDataStorage () ` could return an error message to the hangout chat. I am not sure though, you should check with Marek or Vector.

src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
miasya: @SydneyCaulfeild  All the examples I've seen for initializing service accounts have not closed the stream so I think you're good!

tsa87: Taken from Effective Java, `try-with-resources close` FileInputStream automatically.
```// try-with-resources - the the best way to close resources!
static String firstLineOfFile(String path) throws IOException {
    try (BufferedReader br = new BufferedReader(
        new FileReader(path))) {
            return br.readLine();
       }
} 
```

src/main/java/com/google/flourbot/macro_classes/Macro.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    public String getMacroName () {
+        return macroName;
+    }
+
+    public String getDocId () {
+        return docId;
+    }
+}
```
miasya, line 29: @SydneyCaulfeild I like this class-- thinking ahead!! Macro Object will be useful later on

One question - I'm unsure if we need setters, since initializing them with the constructor should be sufficient. Do we anticipate needing to change them? I remember Marek talking about avoiding having public setter methods since we don't want the wrong things to be able to modify the object (in a code review for week 5). Maybe ask him to double check?

miasya: @lipczak-g

src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
miasya: Good question! Not sure either

src/main/java/com/google/flourbot/macro_classes/Macro.java
tsa87: Good question! fwiw I had my Macro constructor taking a Trigger object and an Action object.

tsa87 (approved):

src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
SydneyCaulfeild: Ok thanks!

src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
SydneyCaulfeild: Ok thanks, I will likely use your suggestion Tony and just propagate the error to the server by throwing an exception from the constructor. @Vector-S @lipczak-g is this an acceptable strategy?

src/main/java/com/google/flourbot/macro_classes/Macro.java
SydneyCaulfeild: Thanks for the advice, and yes Marek said we don't need setters since it should be immutable.

src/main/java/com/google/flourbot/macro_classes/Macro.java
SydneyCaulfeild: Thanks! I've added them to the constructor.

PR #7, opened 2020-07-15
Modify entity module to take Map<String, Object>

tsa87:


tsa87:
@googlebot I signed it!

PR #10, opened 2020-07-15
Server execution

tsa87:
Merge in Execution and Entity module code. Execution module is almost completed, waiting for Spreadsheet Append function.


src/main/java/com/google/flourbot/logic/Executor.java
```diff
@@ -0,0 +1,49 @@
...
+// The Logic class of the server
+public class Executor {
+
+    private final String userEmail;
+    private final FirebaseDataStorage dataStorage;
+    
+    public Executor(String userEmail, FirebaseDataStorage dataStorage) throws Exception {
+        this.userEmail = userEmail;
+        this.dataStorage = dataStorage;
+    }
```
tsa87: CloudSheet should be in constructor, userEmail should not

tsa87: Done

src/main/java/com/google/flourbot/entity/EntityInterface.java
```diff
@@ -0,0 +1,8 @@
+package com.google.flourbot.entity;
+
+import java.lang.Exception;
+import com.google.flourbot.datastorage.FirebaseDataStorage;
+
+public interface EntityInterface {
+    public Macro getMacro(String userEmail, String macroName) throws Exception;
+}
```
tsa87: Interface for Entity Module, which calls converts Map from datastore into a Macro object.

src/main/java/com/google/flourbot/entity/EntityModule.java
```diff
@@ -0,0 +1,86 @@
...
+        Map<String, Object> actionMap = (Map<String, Object>) macroMap.get("action");
+
+        String creatorId = (String) macroMap.get("creatorId");
+        Trigger trigger = getTrigger(triggerMap);
+        Action action = getAction(actionMap);
+
+        Macro macro = new Macro(creatorId, macroName, trigger, action);
+
+        return macro;
+    }
```
tsa87: Implementation of previous interface.

src/main/java/com/google/flourbot/entity/EntityModule.java
```diff
@@ -0,0 +1,86 @@
...
+
+
+    // Singleton Support
+    public static EntityModule getInstance() {
+        if (instance == null) {
+            instance = new EntityModule();
+        }
+        return instance;
+    }
```
tsa87: Entity Module is a singleton class, which means the entire server all uses one single Entity Module.

src/main/java/com/google/flourbot/entity/Macro.java
```diff
@@ -0,0 +1,33 @@
+package com.google.flourbot.entity;
+
+import java.util.Map;
+import java.lang.Object;
+
+import com.google.flourbot.entity.action.Action;
+import com.google.flourbot.entity.trigger.Trigger;
+
+public final class Macro {
```
tsa87, line 9: No major change except naming and changing variable to be private and immutable.

src/main/java/com/google/flourbot/entity/action/ActionType.java
```diff
@@ -0,0 +1,3 @@
+package com.google.flourbot.entity.action;
+
+public enum ActionType { SHEET_APPEND }
```
tsa87, line 4: Enum of all possible action types.

src/main/java/com/google/flourbot/entity/action/SheetAppendAction.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private final String[] columnValue; 
+    private final String sheetAction;
+    private final String sheetUrl;
+
+    public SheetAppendAction(String[] columnValue, String sheetAction, String sheetUrl) {
+        this.columnValue = columnValue;
+        this.sheetAction = sheetAction;
+        this.sheetUrl = sheetUrl;
+    }
```
tsa87, line 19: Specific action class now extends the Action abstract class. Incorporate ActionType Enum to define the type.

src/main/java/com/google/flourbot/entity/trigger/CommandTrigger.java
```diff
@@ -0,0 +1,20 @@
+package com.google.flourbot.entity.trigger;
+
+import java.util.Map;
+
+import com.google.flourbot.entity.trigger.Trigger;
+
+public final class CommandTrigger extends Trigger {
+    
+    private final TriggerType triggerType = TriggerType.COMMAND_TRIGGER;
+    private final String triggerCommand;
```
tsa87, line 10: Same as changes to Action class above

src/main/java/com/google/flourbot/logic/Executor.java
```diff
@@ -0,0 +1,42 @@
...
+
+                break;
+            default: 
+                throw new Exception("Unknown action type");
+        }
+
+        return "Sucessfully executed";        
+    }
+
+}
```
tsa87: Implementation of Execution Class

src/main/java/com/google/flourbot/macro/Action.java
```diff
@@ -1,17 +0,0 @@
-package com.google.flourbot.macro;
```
tsa87, line 1: renamed folder to entity

src/main/java/com/google/flourbot/entity/EntityInterface.java
```diff
@@ -0,0 +1,8 @@
+package com.google.flourbot.entity;
+
+import java.lang.Exception;
+import com.google.flourbot.datastorage.FirebaseDataStorage;
+
+public interface EntityInterface {
+    public Macro getMacro(String userEmail, String macroName) throws Exception;
```
lipczak-g: using generic exception is discouraged because you don't know what can be really thrown. Following up on our discussion today. If something bad can happen that can react to (which is the case here) we should return an optional. If there are any checked exceptions that we cannot react to (e.g. broken db connection) we should turn them into checked exceptions. Please read https://www.geeksforgeeks.org/checked-vs-unchecked-exceptions-in-java/ if you would like to learn more.

lipczak-g: so to be clear, let's return Optional<Macro> here.

src/main/java/com/google/flourbot/entity/EntityInterface.java
```diff
@@ -0,0 +1,8 @@
+package com.google.flourbot.entity;
+
+import java.lang.Exception;
+import com.google.flourbot.datastorage.FirebaseDataStorage;
+
+public interface EntityInterface {
```
lipczak-g: EntityModule

src/main/java/com/google/flourbot/entity/EntityModule.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.flourbot.datastorage.FirebaseDataStorage;
+import com.google.cloud.firestore.QueryDocumentSnapshot;
+import com.google.flourbot.entity.action.Action;
+import com.google.flourbot.entity.action.SheetAppendAction;
+import com.google.flourbot.entity.trigger.Trigger;
+import com.google.flourbot.entity.trigger.CommandTrigger;
+import java.util.Map;
+
+import java.lang.Exception;
+
```
lipczak-g, line 6: EntityModuleImpl is the default naming scheme if you can find something specific like adding Firebase in case of the DataStorage module.

src/main/java/com/google/flourbot/entity/EntityModule.java
```diff
@@ -0,0 +1,86 @@
...
+public class EntityModule implements EntityInterface {
+    
+    private static EntityModule instance = null;
+    private final FirebaseDataStorage datastorage;
+
+    private EntityModule() {
+        this.datastorage = new FirebaseDataStorage();
```
lipczak-g: please read at the bottom how I propose to do it... just wanted to point out explicitly that this type of initialization is risky. DataStorage is supposed to be a singleton, and you have no guarantee that another module will construct it again.

src/main/java/com/google/flourbot/entity/EntityModule.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.flourbot.entity.action.SheetAppendAction;
+import com.google.flourbot.entity.trigger.Trigger;
+import com.google.flourbot.entity.trigger.CommandTrigger;
+import java.util.Map;
+
+import java.lang.Exception;
+
+public class EntityModule implements EntityInterface {
+    
+    private static EntityModule instance = null;
```
lipczak-g: DataStorage as type here and always, if you have a possibility to use interface type - always use the interface ;)

src/main/java/com/google/flourbot/entity/EntityModule.java
```diff
@@ -0,0 +1,86 @@
...
+        return macro;
+    }
+
+    private Trigger getTrigger(Map<String, Object> triggerMap) throws Exception {
+
+        Trigger trigger;        
+        String triggerType = (String) triggerMap.get("type");
+
+        switch(triggerType) {
+            case ("Command Trigger"):
```
lipczak-g: nit, enum would be perfect, but if we cannot have enum we should use the same name standard so "COMMAND_TRIGGER"

src/main/java/com/google/flourbot/entity/EntityModule.java
```diff
@@ -0,0 +1,86 @@
...
+        Trigger trigger;        
+        String triggerType = (String) triggerMap.get("type");
+
+        switch(triggerType) {
+            case ("Command Trigger"): 
+                String command = (String) triggerMap.get("command");
+                trigger = new CommandTrigger(command);
+                break;
+            default:
+                throw new Exception("No Trigger Type found");
```
lipczak-g: IllegalStateException

src/main/java/com/google/flourbot/entity/EntityModule.java
```diff
@@ -0,0 +1,86 @@
...
+        String creatorId = (String) macroMap.get("creatorId");
+        Trigger trigger = getTrigger(triggerMap);
+        Action action = getAction(actionMap);
+
+        Macro macro = new Macro(creatorId, macroName, trigger, action);
+
+        return macro;
+    }
+
+    private Trigger getTrigger(Map<String, Object> triggerMap) throws Exception {
```
lipczak-g: no need for "throws Exception" here and everywhere else.

src/main/java/com/google/flourbot/entity/EntityModule.java
```diff
@@ -0,0 +1,86 @@
...
+        switch(actionType) {
+            case ("Sheet Action"):
+                String[] columnValue = (String[]) actionMap.get("columnValue");
+                String sheetAction = (String) actionMap.get("sheetAction");
+                String sheetUrl = (String) actionMap.get("sheetUrl");
+
+                action = new SheetAppendAction(columnValue, sheetAction, sheetUrl);
+                break;
+            default:
+                throw new Exception("No Action Type found");
```
lipczak-g: IllegalStateException

src/main/java/com/google/flourbot/entity/EntityModule.java
lipczak-g: OK, let's leave it like this for now but we will need to change it later. The proper Spring way of doing it is injection (https://spring.io/guides/gs/multi-module/), but I don't want us to waste time and effort for now. So let's have server initializer class that will be called in the Bot's constructor and it will look something like this:

public static MacroExecutionModule initializeServer() {
  DataStorage dataStorage = new FirestoreDataStorage();
  EntityModule entityModule = new EntityModuleImpl(dataStorage);
  CloudDocClient cloudDocClient = new DriveDocClient();
  MacroExecutionModule  macroExecutionModule = new MacroExecutionModuleImpl(entityModule, cloudDocClient);
  return macroExecutionModule;
}

 This way initializer is the only class that knows about the implementations. Everything else operates on the module interfaces only and no module is initialized outside of the initializer. The main advantages are:
- well defined dependency graph
- cleaner implementations
- security of having at most single instance of each module.

src/main/java/com/google/flourbot/entity/action/ActionType.java
lipczak-g: Add a TODO to make it clear that you plan to expand it.

src/main/java/com/google/flourbot/entity/action/SheetAppendAction.java
```diff
@@ -0,0 +1,36 @@
+package com.google.flourbot.entity.action;
+
+import java.util.Map;
+import com.google.flourbot.entity.action.Action;
+
+
+public final class SheetAppendAction extends Action {
```
lipczak-g, line 7: nit: when you add more actions consider making a parent of this class that extends Action and implements methods like getSheetUrl, which will be common for all sheet actions.

src/main/java/com/google/flourbot/logic/Executor.java
```diff
@@ -0,0 +1,42 @@
...
+import com.google.flourbot.entity.trigger.Trigger;
+import com.google.flourbot.entity.action.Action;
+import com.google.flourbot.entity.EntityModule;
+
+import java.util.Map;
+import java.lang.Exception;
+import com.google.cloud.firestore.QueryDocumentSnapshot;
+
+// The Logic class of the server
+public class Executor {
```
lipczak-g: How about naming it MacroExecutionModuleImpl? And we will need the interface for it.

src/main/java/com/google/flourbot/logic/Executor.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private final EntityModule entityModule;
+    //TODO: Add GoogleSheetHandler here when ready.
```
lipczak-g: please read the initializer comment about that.

src/main/java/com/google/flourbot/logic/Executor.java
```diff
@@ -0,0 +1,42 @@
...
+        Macro macro = entityModule.getMacro(userEmail, macroName);
+        Action action = macro.getAction();
+
+        switch (action.getActionType()) {
+            case SHEET_APPEND:
+                //Document writer stuff
+
+                break;
+            default: 
+                throw new Exception("Unknown action type");
```
lipczak-g: IllegalStateExeception

src/main/java/com/google/flourbot/logic/Executor.java
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        return "Sucessfully executed";
```
lipczak-g: Well we need to do better than this, does it mean that we will need a response action that generates the message for the user? We will face more and more problems like this when we expand the functionality of the bot.

lipczak-g (commented):
Great start! Nice to see everything coming together.

src/main/java/com/google/flourbot/entity/action/ActionType.java
tsa87: Done!

src/main/java/com/google/flourbot/entity/EntityInterface.java
tsa87: Thanks for letting me know about Optional! In the new pull request, null pointer Exceptions are turned into Optional.

src/main/java/com/google/flourbot/entity/EntityInterface.java
tsa87: Done!

src/main/java/com/google/flourbot/entity/EntityModule.java
tsa87: Ah okay, so fooModule is the name for interface and fooModuleImpl is the naming for the actual class

src/main/java/com/google/flourbot/entity/EntityModule.java
tsa87: Good point! Do you have the cast the Interface to Implementation type when calling `getDocument()`?

src/main/java/com/google/flourbot/entity/EntityModule.java
tsa87: Noted! This require some code change on the client side as well.

src/main/java/com/google/flourbot/entity/EntityModule.java
tsa87: Using Java Optional now

src/main/java/com/google/flourbot/entity/EntityModule.java
tsa87: Using Java Optional now

src/main/java/com/google/flourbot/entity/EntityModule.java
tsa87: Done!

src/main/java/com/google/flourbot/logic/Executor.java
tsa87: Using Java Optional now.

src/main/java/com/google/flourbot/entity/EntityModule.java
tsa87: Okay great! I like how all initialization are done only once, and grouped together. I've made this point into an issue for our future iterations.

src/main/java/com/google/flourbot/entity/EntityModule.java
tsa87: Will address in Issue #11

src/main/java/com/google/flourbot/entity/action/SheetAppendAction.java
tsa87: Created #12 for this issue

src/main/java/com/google/flourbot/logic/Executor.java
tsa87: Created #11

src/main/java/com/google/flourbot/logic/Executor.java
tsa87: Created Issue #13

tsa87: Good point, currently this is only capable of returning a plain message, which we might want to expand later.

src/main/java/com/google/flourbot/entity/EntityModule.java
tsa87: Yes I think you can. Changed the type to Interface.

src/main/java/com/google/flourbot/entity/EntityModule.java
tsa87: Issue #14

src/main/java/com/google/flourbot/logic/Executor.java
tsa87: Good idea! An Interface is implemented in the new commit

PR #16, opened 2020-07-17
Server initializer

tsa87:
This pull request have some automatic formatter changes that are not important. Major changes are 1. "Using an initalizer function to construct all the modules we need" and 2. Converting checked exception into the unchecked version where appropriate.

### DIFF
--- a/src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
+++ b/src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
@@ -1,4 +1,4 @@
-package com.google.flourbot.executor;
+package com.google.flourbot.execution;
 
 import com.google.flourbot.entity.Macro;
 import com.google.flourbot.entity.action.Action;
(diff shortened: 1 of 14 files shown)


src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
```diff
@@ -0,0 +1,54 @@
...
+        break;
+      default:
+        throw new IllegalStateException(
+            "Action type named: " + actionType.toString() + "is not implemented yet!");
+    }
+
+    // TODO: Return a response object
+    return "Sucessfully executed";
+  }
+}
```
tsa87, line 54: This pull request have some automatic formatter changes that are not important. Major changes are 1. "Using an initalizer function to construct all the modules we need" and 2. Converting checked exception into the unchecked version where appropriate.

miasya: lgtm!

PR #19, opened 2020-07-18
Pass the baton bot

tsa87:
Not done yet...


lipczak-g (commented):
Thanks for working on it! I don't have comments about dart, but please ask Vector for the review.

flutter_src/lib/logic/state/macro_bloc.dart
```diff
@@ -1,88 +1,135 @@
...
+
+  void preFillCommand() {
+    List<String> variables = actionSheetColumn.value.map((bloc) => "{" + bloc.value + "}" ).toList();
+    String prefill = variables.join(" ");
+    triggerCommand.updateValue(prefill);
   }
 
   @override
   void onSubmitting() async {
     if (state.currentStep == 0) {
```
lipczak-g: nit: Not sure if you have control over it, but enum would seems to be a better solution here: it would be more descriptive and resilient to changes in the order of the steps.

flutter_src/lib/logic/state/macro_bloc_template.dart
```diff
@@ -13,7 +14,16 @@ class FormBlocTemplate {
           formBloc.description.updateValue("Daily Check-in Bot records your teams' daily updates in a Google sheet");
 
           formBloc.actionType.updateValue(Action.SHEET_ACTION);
-          formBloc.actionSheetUrl.updateValue("COPY IN YOUR GOOGLE SHEET URL HERE");
+          formBloc.actionSheetUrl.updateValue("https://docs.google.com/spreadsheets/d/1xF3bB3-rpE5CelBm8TKZ9Qy_D0OSHXNc3YQxcfSvUfc/edit#gid=0");
```
lipczak-g: I assume this is a hack for the demo, but it should be removed before this goes to Master.

flutter_src/lib/logic/state/macro_bloc_template.dart
```diff
@@ -4,6 +4,7 @@ import 'package:macrobaseapp/model/entities/action.dart';
 class FormBlocTemplate {
   static const String DAILY_CHECK_IN = "Daily Check-in";
   static const String FROM_SCRATCH = "Create From Scratch";
+  static const String PASS_THE_BATON = "Pass-the-Baton";
```
lipczak-g, line 7: It is fine for now, but in the long term we want these templates to be stored somewhere, maybe in Firestore? Hardcoding them doesn't make much sense, especially if the data structure is the same as the real macros.

flutter_src/lib/logic/state/macro_bloc_template.dart
```diff
@@ -13,7 +14,16 @@ class FormBlocTemplate {
           formBloc.description.updateValue("Daily Check-in Bot records your teams' daily updates in a Google sheet");
 
           formBloc.actionType.updateValue(Action.SHEET_ACTION);
-          formBloc.actionSheetUrl.updateValue("COPY IN YOUR GOOGLE SHEET URL HERE");
+          formBloc.actionSheetUrl.updateValue("https://docs.google.com/spreadsheets/d/1xF3bB3-rpE5CelBm8TKZ9Qy_D0OSHXNc3YQxcfSvUfc/edit#gid=0");
+        }
+        break;
+      case PASS_THE_BATON:
+        {
+          formBloc.macroName.updateValue("Pass-the-Baton Bot");
```
lipczak-g, line 22: Again, this approach won't scale we need these templates to be stored somewhere and then we will have a general method that populates the fields based on the stored template values.

flutter_src/lib/logic/usecases/login/firebase_auth.dart
```diff
@@ -44,7 +44,7 @@ class FirebaseAuthService {
       final authResult = await _firebaseAuth.signInWithCredential(credential);
       return _userFromFirebase(authResult.user);
     } catch(e) {
-      print(e);
+      return null;
```
lipczak-g: not sure about the context, but returning null from catch doesn't seem right, why not throw instead? I assume we cannot recover from this.

flutter_src/lib/model/entities/action.dart
```diff
@@ -16,7 +16,8 @@ class Action extends Equatable {
 
 // Defines the operation to append a row to a Google Sheet
 class SheetAction extends Action {
-  static const String APPEND_ACTION = "Sheet Append Action";
+  static const String APPEND_ACTION = "Append Action";
```
lipczak-g, line 19: Overall these should not be duplicated between server and UI code because it creates the risk of failures due to typos etc. However, I'm not sure how to solve it in our case, so just pointing this out - no need for action now.

lipczak-g (commented):
I'm fascinated by both dart and your work. Couple high level comments, unfortunately can't do much more given that this is a completely new technology for me. But even at this stage I'm super impressed by the work you've managed to put into it and the quality of the result.

flutter_src/lib/logic/state/macro_bloc_template.dart
tsa87: Done!

flutter_src/lib/logic/state/macro_bloc_template.dart
tsa87: We could store templates in JSON( we also have the function that converts Json to Macro object) However, we still have to implement a function that updates the specific form in our UI using the updateValue() function. I have made this into an issue that we can work on later.

flutter_src/lib/logic/state/macro_bloc_template.dart
tsa87: > We could store templates in JSON( we also have the function that converts Json to Macro object) However, we still have to implement a function that updates the specific form in our UI using the updateValue() function. I have made this into an issue that we can work on later.


Address in #25 issue,

flutter_src/lib/logic/usecases/login/firebase_auth.dart
tsa87: Maybe instead of throw, we can also leave the unchecked (without try/catch) ? Which option do you think is better?

flutter_src/lib/model/entities/action.dart
tsa87: Yea that's a good point. Currently the server code base and UI code base are sort of separate. There isn't a single source of truth for Action/Trigger Type definition. I am not sure how to resolve it currently as well.

flutter_src/lib/logic/state/macro_bloc.dart
tsa87: I will try to look into ENUM in Dart!

flutter_src/lib/logic/state/macro_bloc_template.dart
lipczak-g: Correct, there will still be the need for field-by-field converter, but the values will not be hardcoded.

flutter_src/lib/logic/usecases/login/firebase_auth.dart
lipczak-g: not sure about Dart, but I'm guessing that leaving unchecked is the better option.

PR #23, opened 2020-07-20
Branch for review

SydneyCaulfeild:
This branch was created for our hosts to see changes that have already been merged to master.


src/main/java/com/google/flourbot/api/Bot.java
```diff
@@ -50,7 +50,6 @@
   public static void main(String[] args) {
```
lipczak-g, line 46: L48: MacroExecutionModule macroExecutionModule (don't use implementations outside of the initializer and create descriptive variable names)
L63: nit: I think that "final String replyText;" will force the compiler to make sure the value is set in all paths of the switch.
L83: nit: Do not log PII (wtf/pii) I assume that name is user created name, always use machine generated ids.
L86: This is not a message that the user should see, throw IllegalArgumentException instead.
L96: The file name should be a const like chat scope for now. In the future it should come from some user defined config, so other developers that are using this code can easily customize this part (don't worry about it for now).

src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
```diff
@@ -32,7 +32,6 @@ public FirebaseDataStorage() {
```
lipczak-g, line 48: L5: Don't use * imports, add all classes that you are using explicitly - it is much easier to follow then.
L21: PROJECT_ID, same for L22
L37: nit: field names should be constants as well
L45: nit: return document.exists() ? Optional.of(document) : Optional.empty();
L57: nit: I don't see much of use for this method, I think it can be inlined in the constructor.
L60: The file name should be a constant, see relevant comment in Bot for more details.
L69: nit: return FirestoreClient.getFirestore(); no need for the variable.

src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
```diff
@@ -78,8 +78,8 @@ public EntityModuleImplementation(DataStorage dataStorage) {
```
lipczak-g, line 80: L16: no * imports
L38: nit: there is no need to use the data structure name as suffix for the variable, although this is a common pattern. It is better to call it something like macroData, this way when the object type changes you don't have a problem. This applies to multiple variables in this file and in the project in general, but no need to fix all of them at this point.
L81: nit: columnValues please try to avoid abbreviations.

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
```diff
@@ -38,7 +38,7 @@ public static MacroExecutionModuleImplementation initializeServer() {
   }
```
lipczak-g, line 61: L24: This reminds me we need JavaDoc for each class (and each method in the interface), but the format of it is:
/**
 \* Write doc comment here.
 */
L33: I'd rather have it in a separate file, but size this is the central module now let's keep it here. However you should move the initialization of the DriveClient here as well and pass it as a parameter next to entityModule (make sure to use the interface as type). DriveClient should be a singleton as well.
L42: This should be in a separate method getMacroName(String message, String threadId) this method should implement the caching logic we discussed today.
L55: There is a lot of things hardcoded here which needs be customizable by the macro config. For example, the user should be able to say if they want the timestamp and in which column it should be added.
L60: nit: variable type should be List<String>, this list should be populated in a separate method, this method should take the field config (coming from the macro) as input and it should populate the fields as configured by the user in the UI (we can take care of it after the demo).
L68: DriveClient should be a singleton as mentioned above and the variable name should be cloudDocClient.
L78: The response message should be customizable through UI as well.

src/main/java/com/google/flourbot/api/Bot.java
SydneyCaulfeild: Resolved: L48, L83, L86, L96

SydneyCaulfeild: Resolved: L96

src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
SydneyCaulfeild: Resolved all.

src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
SydneyCaulfeild: Resolved all.

src/main/java/com/google/flourbot/api/Bot.java
SydneyCaulfeild: @lipczak-g when I changed replyText to "final String replyText;" I get an error saying "variable replyText might not have been initialized". I think this is because it is not assigned any value in the "REMOVED_FROM_SPACE" case. I don't think it makes sense to assign anything in this case because the bot has been removed from the room so it can't reply with anything anyways.

lipczak-g: oh, sorry for missing that, in such case the trick will not work. In such case I think we should add is empty check before it is used and throw IllegalArgumentException if it has not been initialized. It is always better to throw an exception instead of returning obviously incorrect result to the caller.

src/main/java/com/google/flourbot/api/Bot.java
```diff
@@ -77,15 +77,13 @@ public void onEvent(@RequestBody JsonNode event) throws IOException, GeneralSecu
...
         break;
       case "REMOVED_FROM_SPACE":
-        String name = event.at("/space/name").asText();
-        logger.info(String.format("Bot removed from %s", name));
+        logger.info("Bot removed from space");
         break;
       default:
-        replyText = "Cannot determine event type";
-        break;
+        throw new IllegalArgumentException();
```
lipczak-g, line 85: Add "Unrecognized event type: "+event.at("/type").asText() as the argument to the exception.

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
lipczak-g: Just leaving a note that these comments have not been addressed yet.

src/main/java/com/google/flourbot/api/Bot.java
miasya: resolved

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
miasya: L42, L55, L60, L68 resolved

src/main/java/com/google/flourbot/api/Bot.java
SydneyCaulfeild: Update: resolved all.

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
SydneyCaulfeild: Resolved: L33

PR #24, opened 2020-07-22
add readRange, readRow, etc

miasya:
…for our sheet needs. minor changes involving making some variables constants and condensing extra lines. remove createNew()


src/main/java/com/google/flourbot/api/DriveCloudSheet.java
```diff
@@ -12,49 +12,66 @@
 import java.util.List;
 
 public class DriveCloudSheet implements CloudSheet {
-  Sheets sheetsService;
-  String spreadsheetId;
+  private String valueInputOption = "USER_ENTERED";
```
lipczak-g: example of a proper constant in Java:
private static final String VALUE_INPUT_OPTION = "USER_ENTERED";

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
```diff
@@ -12,49 +12,66 @@
...
 
-  // TODO: replace hardcoded range with calculated range (so far values must be length 3)
-  public void appendRow(List<String> values) throws IOException, GeneralSecurityException {
+  private String getRange(int numEntryColumns, int numSkipColumns) {
+    // Returns range for sheets request by calculating the alphabetic name of
+    // the start and end column based on the number of columns filled and pre-skipped
+
+    String startColumn = this.toAlphabetic(numSkipColumns);
+    String endColumn = this.toAlphabetic(numSkipColumns + numEntryColumns - 1);
+    return "Sheet1!" + startColumn + "1:" + endColumn + "1";
```
lipczak-g: This looks like the job for String.format(), here you have an example and some context: https://dzone.com/articles/java-string-format-examples

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
```diff
@@ -12,49 +12,66 @@
...
     this.sheetsService = documentService;
   }
 
-  // TODO: replace hardcoded range with calculated range (so far values must be length 3)
-  public void appendRow(List<String> values) throws IOException, GeneralSecurityException {
+  private String getRange(int numEntryColumns, int numSkipColumns) {
+    // Returns range for sheets request by calculating the alphabetic name of
+    // the start and end column based on the number of columns filled and pre-skipped
+
+    String startColumn = this.toAlphabetic(numSkipColumns);
```
lipczak-g: no need for "this."

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
```diff
@@ -12,49 +12,66 @@
...
+    // Recursive strategy to build a string that represents the alphabetic name
+    // of a column based on column number i
+    // e.g. 0 -> A, 1 -> B... 25 -> Z, 26 -> AA...
     
-    // Generate request to append to sheet
-    String range = "Sheet1!A1:C1";
-    String valueInputOption = "USER_ENTERED";
-    String insertDataOption = "INSERT_ROWS";
+    // Only consider the absolute value of the number
+    if (i < 0) {
```
lipczak-g, line 45: https://en.wikipedia.org/wiki/Fail-fast

Input doesn't make sense? IllegalArgumentException, also Math.abs() is the right way of doing it.

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
```diff
@@ -12,49 +12,66 @@
...
+      i *= -1;
+    }
+
+    // Every calculation produces 1 letter
+    int quotient = i / 26;
+    int remainder = i % 26;
+    char letter = (char)((int)'A' + remainder);
```
lipczak-g: I believe there are some errors in your logic, and also it can be written in more concise way, maybe something like this
String firstLetter = ""+(char)(i % 27 + 'A'); // putting the number first should make Java cast the letter automatically
return firstLetter + (i > 26 ? (char)((i-1) / 26 + 'A') : "")

Not sure about that, so please check it, also, this is a perfect method to add a test for.

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
miasya: I like this so much better!

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
miasya: Whoops that's a silly mistake - thanks for pointing it out!

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
miasya: When do we include or omit "this" ? What are the best practices

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
miasya: I like the brevity, but I included the function so that later on we could have reading methods that could accept both a number or a value.
What's the tradeoff of an implicit vs explicit cast?

miasya: I'm having trouble with (char)(i % 27 + 'A') -> gives a runtime error so I think I'll leave it for now

src/main/java/com/google/flourbot/api/CloudDocClient.java
```diff
@@ -4,7 +4,5 @@
 import java.security.GeneralSecurityException;
 
 public interface CloudDocClient {
-
-  // TODO: change exception handling to try catch. throw illegal state exception
```
lipczak-g: that's a good comment, we don't want to throw these.

src/main/java/com/google/flourbot/api/CloudSheet.java
```diff
@@ -8,4 +8,11 @@
 
 public interface CloudSheet {
   public void appendRow(List<String> values) throws IOException, GeneralSecurityException;
+
+  public List<List<String>> readRange(String range);
+
+  public List<String> readRow(int row);
+  public List<String> readCol(int column);
```
lipczak-g: readColumn, full names are easier to understand.

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
```diff
@@ -10,51 +10,104 @@
...
+      throw new IllegalArgumentException(String.format("Column number %d is invalid", i));
+    }
+
+    // Every calculation produces 1 letter
+    int quotient = i / 26;
+    char letter = (char)(i % 26 + 'A');
+    if (quotient == 0) {
+        return "" + letter;
+    } else {
+        return toAlphabetic(quotient-1) + letter;
```
lipczak-g, line 55: recursion - nice! And it should work for more than 2 letters (not even sure it is possible in Spreadsheet, but still nice)

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
```diff
@@ -10,51 +10,104 @@
...
-    String valueInputOption = "USER_ENTERED";
-    String insertDataOption = "INSERT_ROWS";
+    // Negative numbers cause errors
+    if (i < 0) {
+      throw new IllegalArgumentException(String.format("Column number %d is invalid", i));
+    }
+
+    // Every calculation produces 1 letter
+    int quotient = i / 26;
+    char letter = (char)(i % 26 + 'A');
```
lipczak-g, line 51: I still think this should be 27, can you convince me?

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
```diff
@@ -10,51 +10,104 @@
...
+    if (quotient == 0) {
+        return "" + letter;
+    } else {
+        return toAlphabetic(quotient-1) + letter;
+    }
+  }
+
+  public List<List<String>> readRange(String range) {
```
lipczak-g, line 59: What is this range and who defines it? If the format is A1:B3, it would be good to expose a method that builds this range for the caller if needed. Or have the method that takes startRow, endRow, startColumn, endColumn and build the range, similar to what you do with getColumn. Having a method that validates the range would be good as well. But all of these are good to have, no need to implement now.

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
```diff
@@ -10,51 +10,104 @@
...
+    List<List<String>> values = readRange(String.format("Sheet1!%d:%d", row, row));
+
+    if (values != null || !values.isEmpty()) {
+      return (List<String>) values.get(0);
+    } else {
+      return Collections.emptyList();
+    }
+  }
+
+  public List<String> readCol(int column) {
```
lipczak-g: I like this design, you are building a generic interface, so accepting both int and String as inputs is forward looking.

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
```diff
@@ -10,51 +10,104 @@
...
+    // Overload readCol to take a number
+    return readCol(toAlphabetic(column));
+  }
+
+  public List<String> readCol(String column) {
+    List<List<String>> values = readRange(String.format("Sheet1!%s:%s", column, column));
+
+    if (values != null || !values.isEmpty()) {
```
lipczak-g, line 84: should be && not ||

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
```diff
@@ -10,51 +10,104 @@
...
+    }
+  }
 
-    ValueRange requestBody = new ValueRange();
+  public List<List<String>> readSheet(String sheetName) {
+    return readRange(sheetName);
+  }
+
+  public void appendRow(List<String> values) throws IOException, GeneralSecurityException {
+    // Method generates and sends request to append rows to bottom of sheet containing values
```
lipczak-g, line 96: nit: This is should be javadoc in the interface.

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
```diff
@@ -66,15 +67,25 @@ public String execute(String userEmail, String message) throws IOException, Gene
...
         DriveClient cdc = new DriveClient();
-        CloudSheet cs = cdc.getCloudSheet(documentId);
-        cs.appendRow(values);
+        CloudSheet cloudSheet = cdc.getCloudSheet(documentId);
+        cloudSheet.appendRow(values);
+        break;
+      case SHEET_READ_ROW:
+        // TODO: after merge with master, move CloudSheet init to before the switch statement
+        CloudSheet cloudSheet = cloudDocClient.getCloudSheet(documentId);
+        List<String> sheetData = cloudSheet.readRow(2); //TODO: Replace hardcoded
```
lipczak-g, line 137: normally I'd say it is better to leave this change for later when we have the proper value from the parameter, but it's ok to merge to master now that you have it I guess.

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
```diff
@@ -66,15 +67,25 @@ public String execute(String userEmail, String message) throws IOException, Gene
...
+        CloudSheet cloudSheet = cloudDocClient.getCloudSheet(documentId);
+        List<String> sheetData = cloudSheet.readCol("B");//TODO: Replace hardcoded
         break;
       default:
         throw new IllegalStateException(
             "Action type named: " + actionType.toString() + "is not implemented yet!");
     }
 
     // TODO: Return a response object
+    // If read row or read col - give the row or col
```
lipczak-g, line 149: I assume we may want to implement some post processing as well, e.g. formatting with commas.

lipczak-g (approved):

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
miasya: A->Z is 26 letters, so if we do i % 26, we have results [0,25] which is 26 different numbers. This gives us A + 0 = A... A + 25 = Z

src/main/java/com/google/flourbot/api/CloudDocClient.java
miasya: Added as issue so that I remember!

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
miasya: Can values be both null and empty at the same time?

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
miasya: Does that mean to add the method description to the CloudSheet.java?

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
miasya: Yes!

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
miasya: Ah I like the idea of having a method for it. One option is just to hope the user uses A1 notation but I think that causes more friction on their end so we should handle it

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
lipczak-g: Sorry, yeah I got confused because I assumed that the columns will be indexed from 1... it is still not clear to me that they won't, so it would be good to make it clear in javadoc of all the methods that use it.

PR #28, opened 2020-07-23
Remove duplicate column logic

miasya:
Tony and I both made structures for controlling the input of each column based on type, here, I remove duplicate logic but keep the new date formatting function


miasya:
Ok I don't know where the merge commits came from - but there's only 1 file changed so I think we're fine

PR #29, opened 2020-07-23
Remove "@MacroBot macroName" from message.

SydneyCaulfeild:
Only write the message to the "Content" column in the spreadsheet. Can see before and after: https://screenshot.googleplex.com/NTUKV6HRxTn.png.

### DIFF
--- a/src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
+++ b/src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
@@ -77,6 +77,10 @@ public class MacroExecutionModuleImplementation implements MacroExecutionModule 
       case SHEET_APPEND:
         // Read instructions on what to write
 
+        // Remove "@Macrobot macroName" from message
+        String[] messages = message.split(" ", 3);
+        message = messages[2];
+
         SheetEntryType[] columns = ((SheetAppendAction) action).getColumnValue();
         // Prepare values to write into the sheet
         ArrayList<String> values = new ArrayList<String>();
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
```diff
@@ -77,6 +77,10 @@ public String execute(String userEmail, String message, String threadId) throws
       case SHEET_APPEND:
         // Read instructions on what to write
 
+        // Remove "@Macrobot macroName" from message
+        String[] messages = message.split(" ", 3);
```
lipczak-g: That's a good start but we are missing:
- error handling, what if the message contains less than 3 sections, what if the user made a mistake and used two spaces etc.
- two sections are possible if the message is missing the macro name (i.e. the cached value is used).

lipczak-g: Feel free to submit and we can come back to the details later if there is time.

lipczak-g (approved):

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
tsa87: There might be " " in the user's message. Ie "@Marcobot macroName Good Morning" becomes ["@MacroBot", "macroName", "Good", "Morning"] 

In this case, the array should be sliced from [2 : length], and rejoined with " ".

SydneyCaulfeild: @tsa87 I believe the split I have will still work (take a look at this screenshot: https://screenshot.googleplex.com/wGk1Qj8FUhb.png). However, the bot's name will have to be one word without spaces, so we should maybe specify that in the website when users create their macro.

SydneyCaulfeild: @lipczak-g I have added an error check and an if statement to split the message differently depending on whether or not the message contains the macroname.

tsa87: Okay great! If you have a forloop, that works.

PR #30, opened 2020-07-24
Add Email Scope to specify the people who can use the bot.

tsa87:
Works, however deprecates previous macros created without the `Scope` field.

### DIFF
--- a/src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
+++ b/src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
@@ -18,7 +18,7 @@ public class FirebaseDataStorage implements DataStorage {
   private static final String PROJECT_ID = "stepladder-2020";
   private static final String SERVICE_ACCOUNT = "/key.json";
   private static final String COLLECTION_NAME = "macros";
-  private static final String USER_IDENTIFIER = "creatorId";
+  private static final String SCOPE_IDENTIFIER =  "scope";
   private static final String MACRO_IDENTIFIER = "macroName";
 
   public FirebaseDataStorage() {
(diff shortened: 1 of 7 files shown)


miasya (approved):
I like that it uses email scopes - is the next step to add the saving names in added to space?

src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
```diff
@@ -45,8 +45,8 @@ public FirebaseDataStorage() {
     // Create a query to find a macro named macroName belonging to userEmail
     Query query =
         db.collection(COLLECTION_NAME)
-            .whereEqualTo(USER_IDENTIFIER, userEmail)
-            .whereEqualTo(MACRO_IDENTIFIER, macroName);
+            .whereEqualTo(MACRO_IDENTIFIER, macroName)
+            .whereArrayContains(SCOPE_IDENTIFIER, userEmail);
```
SydneyCaulfeild, line 49: I'm not sure if using an index helps or speeds the query up, but worth looking into it (https://firebase.corp.google.com/project/stepladder-2020/database/firestore/indexes).

src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
```diff
@@ -59,7 +59,7 @@ public FirebaseDataStorage() {
       document = querySnapshot.get().getDocuments().get(0);
       return document.exists() ? Optional.of(document) : Optional.empty();
     } catch (IndexOutOfBoundsException e) {
-      throw new IllegalStateException(e);
+      return Optional.empty();
```
SydneyCaulfeild, line 62: Thanks for finding this!!

SydneyCaulfeild (approved):

src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
lipczak-g, line 62: Oh, if IndexOutOfBoundsException is possible, we should check isEmpty first before .get(0) and return empty optional if that's the case. Exceptions are expensive and complicated, if we can we should avoid them.

tsa87: Oh I see. I don't see a .emtpy() method on `DocumentSnapshot`  type though.

PR #33, opened 2020-07-24
add card response and build support for new read functions in other modules

miasya:
- Fix list casting issues and extra imports (and removing star imports)
- add SHEET_READ_ROW object
- add SHEET_READ_ROW,  SHEET_READ_COLUMN, SHEET_READ_SHEET enums and cases in switch statement with different String replies for each case
- rename SheetAppendAction.java to SheetAppendRowAction.java
- add CardResponse.java class that holds the String reply for different cases and displays it in the new card format with basic HTML (tables not supported unfortunately)
- change Bot.java and MacroExecutionModule.java to use the new CardResponse object
- add misc. extra exception handling logic


miasya:
SheetAppendAction.java and SheetReadRowAction.java are currently nearly identical, aside from the ActionType field.

In the future, should we just create one class called SheetAction.java that supports all sheet actions (read/write/delete?).
A concern that was brought up is that some fields will be NULL if they aren't relevant to the actionType.

But also, these actions have a lot of overlap, and adding different classes for several very similar actions takes up a lot more lines of code (not only to make the classes, but also to handle the different classes in our other modules with switch statements and casting and all that).

What are everyone's thoughts?

src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
```diff
@@ -91,7 +92,23 @@ public EntityModuleImplementation(DataStorage dataStorage) {
         String sheetAction = (String) actionData.get("sheetAction");
```
tsa87, line 92: Should convert to `ActionType` ENUM using `ActionType.valueOf(sheetAction);`

tsa87 (approved):
Looks Great!

src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
miasya: Are we changing the representation in the DB? Right now it's still "Sheet Append Action" but if we change it to "SHEET_APPEND_ROW" then I think the casting is a great idea! Saves us the extra code

miasya:
So it was working at [this commit](https://github.com/googleinterns/step81-2020/pull/33/commits/1bba7fe9a84b478c56ee9415ac7eb92fd7ef375b) but then after I redeployed I keep getting "No macro of name: CongraBot found" so I'm wondering if there's an issue somewhere with that

tsa87:
> So it was working at [this commit](https://github.com/googleinterns/step81-2020/pull/33/commits/1bba7fe9a84b478c56ee9415ac7eb92fd7ef375b) but then after I redeployed I keep getting "No macro of name: CongraBot found" so I'm wondering if there's an issue somewhere with that

Might be because since the last commit, calling a macro requires the your email to be in the `scope` list attribute. If that is the reason, you can find an example macro in firebase that has the field. 

- But I believe we are gonna use the Map<room, owner> approach soon

src/main/java/com/google/flourbot/api/Bot.java
```diff
@@ -67,26 +71,31 @@ public void onEvent(@RequestBody JsonNode event) throws IOException, GeneralSecu
...
       case "MESSAGE":
         // Sends request to execution module
         String email = event.at("/message/sender/email").asText();
         String message = event.at("/message/text").asText();
         String threadId = event.at("/message/thread/name").asText();
-        replyText = macroExecutionModule.execute(email, message, threadId);
+        
+        ChatResponse chatResponse = macroExecutionModule.execute(email, message, threadId);
+        replyText = chatResponse.getReplyText();
+        Card card = CardResponse.createCardResponse(replyText);
```
lipczak-g, line 83: Are we sure that we want everything to be a card, is a card with just a text the same as simple text response?

src/main/java/com/google/flourbot/api/CardResponse.java
```diff
@@ -0,0 +1,96 @@
+/*
```
lipczak-g, line 1: this should be javadoc, so starts with /** and the comment should be immediately above the class definition.

src/main/java/com/google/flourbot/api/CardResponse.java
```diff
@@ -0,0 +1,96 @@
...
+
+  private static final String INTERACTIVE_TEXT_BUTTON_ACTION = "doTextButtonAction";
+  private static final String INTERACTIVE_IMAGE_BUTTON_ACTION = "doImageButtonAction";
+  private static final String INTERACTIVE_BUTTON_KEY = "originalMessage";
+  private static final String HEADER_IMAGE = "https://shop.googlemerchandisestore.com/store/20160512512/assets/items/largeimages/GGOEACBA116699.jpg";
+  private static final String BOT_NAME = "MacroBot";
+  private static final String REDIRECT_URL = "https://goo.gl/kwhSNz"; // TODO: Change to spreadsheet link
+
+  /**
+    * Creates a card-formatted response based on the message sent in Hangouts Chat.
```
lipczak-g, line 46: Nice javadoc, which makes me think this was copied from somewhere - that's prefectly fine, but it would be nice to call it out for our internal knowledge.

src/main/java/com/google/flourbot/api/CardResponse.java
```diff
@@ -0,0 +1,96 @@
...
+
+    // TEXT PARAGRAPH
+    TextParagraph textParagraphWidget = new TextParagraph().setText("<b>The reply text we got:</b> is <br></br>" + message);
+    widgets.add(new WidgetMarkup().setTextParagraph(textParagraphWidget));
+
+    // TODO: Add cute image if CongraBot (do this after we make a response object)
+
+    // KEY VALUE
+    KeyValue keyValueWidget = new KeyValue()
+                  .setTopLabel("KeyValue widget")
```
lipczak-g, line 74: Please don't submit test strings. It is better to leave things out of the initial submission rather that adding something that is unfinished and risking getting it to production.

src/main/java/com/google/flourbot/api/CardResponse.java
```diff
@@ -0,0 +1,96 @@
...
+                  .setContent("This is a KeyValue widget")
+                  .setBottomLabel("The bottom label")
+                  .setIcon("DESCRIPTION");
+    widgets.add(new WidgetMarkup().setKeyValue(keyValueWidget));
+
+    // TEXT BUTTON
+    OpenLink openLink = new OpenLink().setUrl(REDIRECT_URL);
```
lipczak-g, line 81: There is a lot of things added here, it seems that we need some conditions to apply only relevant fields, correct?

src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
```diff
@@ -91,7 +92,23 @@ public EntityModuleImplementation(DataStorage dataStorage) {
...
+        switch(sheetAction) {
+          case ("Sheet Append Action"):
+            action = new SheetAppendRowAction(columnValue, sheetAction, sheetUrl);
+            break;
+ 
+          case ("Read Row Action"):
+            action = new SheetReadRowAction(columnValue, sheetAction, sheetUrl);
+            break;
+
+          case ("Batch Action"):
```
lipczak-g, line 106: what is going on whith Batch Action? It is not safe to leave it like this. Throw java.lang.UnsupportedOperationException instead to make sure we come back to this code to fix it.

src/main/java/com/google/flourbot/entity/action/sheet/SheetAppendRowAction.java
```diff
@@ -0,0 +1,49 @@
+package com.google.flourbot.entity.action.sheet;
```
lipczak-g, line 1: btw, It would be nice to separate these two changes, because they address completely different functionality but that's ok for now.

src/main/java/com/google/flourbot/entity/action/sheet/SheetReadRowAction.java
```diff
@@ -6,15 +6,15 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-public final class SheetAppendAction extends Action {
+public final class SheetReadRowAction extends Action {
```
lipczak-g, line 9: The fact that it is so easy to change this method shows that it would be very useful to have parent SheetAction that gathers the common methods for all Sheet actions - don't worry about it for now though.

src/main/java/com/google/flourbot/execution/ChatResponse.java
```diff
@@ -0,0 +1,130 @@
...
+ */
+
+package com.google.flourbot.execution;
+
+import com.google.flourbot.entity.action.ActionType;
+
+import java.util.Collection;
+import java.util.List;
+
+public class ChatResponse {
```
lipczak-g, line 12: This class and its methods could really use some javadoc because it is not clear what is the goal here... one thing to consider is that for the simplest responses we may want to simply return the text.

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
```diff
@@ -134,23 +137,31 @@ public String execute(String userEmail, String message, String threadId) throws
...
         break;
+
       case SHEET_READ_COLUMN:
         List<String> columnData = cloudSheet.readColumn("B");//TODO: Replace hardcoded
-        replyText = columnData.get(0);
+        chatResponse = ChatResponse.createChatResponseWithList(columnData, actionType, documentUrl);
+        break;
+
+      case SHEET_READ_SHEET:
+        List<List<String>> sheetData = cloudSheet.readSheet("Sheet1");
```
lipczak-g, line 154: I guess this can be our default, but it would be good to give the user a chance to define the sheet name when the macro is created.

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
```diff
@@ -159,4 +170,9 @@ private String getDate(String pattern) {
     DateTimeFormatter myFormatObj = DateTimeFormatter.ofPattern(pattern);
     return myDateObj.format(myFormatObj);
   }
+
+  private String selectRandomFromData(List<String> data) {
```
lipczak-g, line 174: Don't add private methods if they are not used yet, we can do it in the next change.

lipczak-g (commented):
Nice work - a lot is done here.

src/main/java/com/google/flourbot/api/CardResponse.java
miasya: What does "call it our in our internal knowledge" mean in practice? I mentioned it was from the sample code Sydney sent in the commit message - but where should I put it next time?

src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
tsa87: Oh good point! The firestore representation is not all caps.

src/main/java/com/google/flourbot/execution/ChatResponse.java
miasya: Got it!

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
miasya: Yes! My only concern is that the user can edit the sheet name afterwards and we won't see the change. We would have to add more logic to find all the sheetnames in the Sheets doc and I don't know how we would know to select the proper one. So it seems simpler to tell the user to keep the default name

miasya: At least for our goals, we can come back to this if time permits

src/main/java/com/google/flourbot/api/CardResponse.java
miasya: removed

src/main/java/com/google/flourbot/api/Bot.java
miasya: Updated! In PR #36

PR #34, opened 2020-07-24
Pass the baton 2

tsa87:
Allow form to handle more complex sheet actions, build classes.

### DIFF
--- a/src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
+++ b/src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
@@ -18,7 +18,7 @@ public class FirebaseDataStorage implements DataStorage {
   private static final String PROJECT_ID = "stepladder-2020";
   private static final String SERVICE_ACCOUNT = "/key.json";
   private static final String COLLECTION_NAME = "macros";
-  private static final String USER_IDENTIFIER = "creatorId";
+  private static final String SCOPE_IDENTIFIER =  "scope";
   private static final String MACRO_IDENTIFIER = "macroName";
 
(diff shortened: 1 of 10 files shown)


miasya (approved):

src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
```diff
@@ -45,8 +45,8 @@ public FirebaseDataStorage() {
     // Create a query to find a macro named macroName belonging to userEmail
     Query query =
         db.collection(COLLECTION_NAME)
-            .whereEqualTo(USER_IDENTIFIER, userEmail)
-            .whereEqualTo(MACRO_IDENTIFIER, macroName);
+            .whereEqualTo(MACRO_IDENTIFIER, macroName)
+            .whereArrayContains(SCOPE_IDENTIFIER, userEmail);
```
SydneyCaulfeild, line 49: Once again, you may want to create a composite index on Firebase.

SydneyCaulfeild (approved):

src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
tsa87: Sure, if we aren't going with using the RoomID, I will add the index in.

lipczak-g, line 49: do we want to allow the case in which the user leaves scope empty and then we allow all users? Or in this case only the user should be able to use the macro? Maybe we can use some kind of keywords for it like "ALL" or "PERSONAL"

PR #35, opened 2020-07-24
Add a hashmap to store creator with a room.

SydneyCaulfeild:


src/main/java/com/google/flourbot/api/Bot.java
```diff
@@ -69,12 +72,28 @@ public void onEvent(@RequestBody JsonNode event) throws IOException, GeneralSecu
...
-        replyText = macroExecutionModule.execute(email, message, threadId);
+
+        String[] words = message.split(" "); 
+
+        if (words[1].equalsIgnoreCase("initiate")) {
+            String creator = event.at("/message/sender/email").asText();
+            roomToCreator.put(event.at("/space/name").asText(), creator);
+            replyText = "Your macros have now been initiated for this room. All users in this room can use your macros.";
+        }
+        else {
```
lipczak-g: if {
} else if {
} else {
}

src/main/java/com/google/flourbot/api/Bot.java
```diff
@@ -69,12 +72,28 @@ public void onEvent(@RequestBody JsonNode event) throws IOException, GeneralSecu
...
         String message = event.at("/message/text").asText();
         String threadId = event.at("/message/thread/name").asText();
-        replyText = macroExecutionModule.execute(email, message, threadId);
+
+        String[] words = message.split(" "); 
+
+        if (words[1].equalsIgnoreCase("initiate")) {
+            String creator = event.at("/message/sender/email").asText();
+            roomToCreator.put(event.at("/space/name").asText(), creator);
+            replyText = "Your macros have now been initiated for this room. All users in this room can use your macros.";
```
lipczak-g: There is some verification needed here. We don't know if the person who initiates the macro is in fact the creator, or if the macro even exists.

src/main/java/com/google/flourbot/api/Bot.java
```diff
@@ -69,12 +72,28 @@ public void onEvent(@RequestBody JsonNode event) throws IOException, GeneralSecu
...
+        }
+        else {
+            if (roomToCreator.containsKey(event.at("/space/name").asText())) {
```
lipczak-g: OK, let's assume that the creator called initiate properly before, how do we know that this is the same macro that is being called here. I think the map should be keyed on the room, macro name pair. You can combine them into a single string.

src/main/java/com/google/flourbot/api/Bot.java
```diff
@@ -69,12 +72,28 @@ public void onEvent(@RequestBody JsonNode event) throws IOException, GeneralSecu
...
+                String email = roomToCreator.get(event.at("/space/name").asText());
+                replyText = macroExecutionModule.execute(email, message, threadId);
+            }
+            else {
+                replyText = "You do not have access to this macro.";
+            }
+        }
         break;
       case "REMOVED_FROM_SPACE":
+        roomToCreator.remove(event.at("/space/name").asText());
```
lipczak-g: Hmmm... if we go with the room_macro as the key, the cleanup will be harder. Maybe the proper structure is Map<String, Map<String, String>> - first the room as the key, then the macro name.

SydneyCaulfeild:
**Key points:**
- The method getMacroName() is now called from Bot.java instead of in the execute() function in MacroExecutionModuleImplementation.java. The execute() function now accepts the MacroName as a parameter. This change was made because Bot.java needs the macro's name to store it in the map.
- The execute() function now accepts two emails instead of one. The first, userEmail, is the message sender's email which is needed to correctly log who sent what message in the spreadsheet. The second, macroCreatorEmail, is the email of the macro's creator. This is needed to perform the query to get the macro.
- This branch will NOT work with the new scope field that Tony added where users give a list of emails with permission to the bot in the website because this approach uses different fields in the database query.

src/main/java/com/google/flourbot/api/Bot.java
```diff
@@ -57,26 +64,75 @@ public static void main(String[] args) {
...
+
       case "ADDED_TO_SPACE":
         String spaceType = event.at("/space/type").asText();
         if ("ROOM".equals(spaceType)) {
           String displayName = event.at("/space/displayName").asText();
           replyText = String.format("Thanks for adding me to %s. Type \"@MacroBot help\" at any time to see my instructions.", displayName);
+          helpMessage = "To use a room member's macro, the creator of a macro must write \"@MacroBot initiate MacroName \". " + 
+            "Once this has been sent, any room member can use the initialized macro by writing \"@MacroBot MacroName <your message> \" " + 
+            "If your macro has already been used in a thread, you may omit the MacroName and can simply write \"@MacroBot <your message>\".";
+
```
miasya, line 70: Could we make this help message a private static final constant?

src/main/java/com/google/flourbot/api/Bot.java
```diff
@@ -57,26 +64,75 @@ public static void main(String[] args) {
...
+                String userEmail = event.at("/message/sender/email").asText();
+                replyText = macroExecutionModule.execute(userEmail, macroCreatorEmail, message, threadId, macroName);
+            }  
+            else {
+                replyText = "You do not have any macros initialized for this room.";
+            } 
+        }
+        else {
+            replyText = "That is not a valid message.";
+        }
```
miasya: I like this! I think it would be good to separate it into a different function starting at line 86? What do you think

src/main/java/com/google/flourbot/execution/MacroExecutionModule.java
```diff
@@ -4,5 +4,5 @@
 import java.security.GeneralSecurityException;
 
 public interface MacroExecutionModule {
-  String execute(String userEmail, String message, String threadId) throws IOException, GeneralSecurityException;
+  String execute(String userEmail, String macroCreatorEmail, String message, String threadId, String macroName) throws IOException, GeneralSecurityException;
```
miasya: +1 this is more specific as to what the emails refer to

src/main/java/com/google/flourbot/api/Bot.java
miasya: also in our demo - probably important to have screenshots illustrating the space initalization step and the thread step as two separate things. Could be confusing to a new user, and if we make screenshots of the UI I think people will understand it better

miasya (approved):
Move the message splitting logic into a separate function then we're good to merge!

src/main/java/com/google/flourbot/api/Bot.java
```diff
@@ -26,20 +26,27 @@
...
+import java.util.LinkedList;
 
 @SpringBootApplication
 @RestController
 public class Bot {
+
   static final String CHAT_SCOPE = "https://www.googleapis.com/auth/chat.bot";
   private static final String SERVICE_ACCOUNT = "/service-acct.json";
   private static final Logger logger = Logger.getLogger(Bot.class.getName());
+  private static String helpMessage;
```
tsa87: Add the constant definition here

src/main/java/com/google/flourbot/api/Bot.java
SydneyCaulfeild: @miasya I actually get an error saying "cannot assign a value to final variable helpMessage because I am defining it differently in different places.

src/main/java/com/google/flourbot/api/Bot.java
SydneyCaulfeild: I agree. Would you keep this function in Bot.java or put it in the execution class? @tsa87 any thoughts on this?

src/main/java/com/google/flourbot/api/Bot.java
```diff
@@ -57,26 +64,75 @@ public static void main(String[] args) {
...
+        // Reply to the help message.
+        if (words[1].equalsIgnoreCase("help")) {  
+            replyText = helpMessage; 
+        }        
+        else if (words.length >= 2) {
+            // If a macro is being initialized for a room.
+            if (words[1].equalsIgnoreCase("initiate")) {
+                // getMacroName() requires the macroName to be the second word of the message, so remove the word "initiate". This will put the macro name as the second word in the message.
+                List<String> wordsWithoutInitiate = new LinkedList(Arrays.asList(words));
+                wordsWithoutInitiate.remove("initiate");
```
tsa87: If word 1 is indeed "initiate",  taking and joining the sub array from index [2: end] might be easier.

tsa87 (approved):
LGTM

src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
```diff
@@ -46,7 +46,7 @@ public FirebaseDataStorage() {
     Query query =
         db.collection(COLLECTION_NAME)
             .whereEqualTo(MACRO_IDENTIFIER, macroName)
-            .whereArrayContains(SCOPE_IDENTIFIER, userEmail);
+            .whereEqualTo(CREATOR_IDENTIFIER, userEmail);
```
tsa87, line 49: Are we not using the list scope anymore?

src/main/java/com/google/flourbot/api/Bot.java
SydneyCaulfeild: What do you mean by constant definition? I don't believe it can be final because it is defined differently depending on if the bot is used in a DM or a room.

src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
SydneyCaulfeild: I am not sure, personally I don't know how to make them work together. Also, if we have the room sharing feature working, I am not sure what capabilities the scope field adds. What do you think?

src/main/java/com/google/flourbot/api/Bot.java
SydneyCaulfeild: Could you please elaborate on how this is done? I am looking to remove the first element of the String array but keep everything else.

src/main/java/com/google/flourbot/api/Bot.java
lipczak-g: helpMessage should be a local variable (in the method) whenever you set it it should be something like:
helpMessage = ADD_TO_SPACE_MESSAGE;
where ADD_TO_SPACE_MESSAGE is a constant defined at the class level.

lipczak-g: same about replyText btw.

src/main/java/com/google/flourbot/api/Bot.java
```diff
@@ -57,26 +64,76 @@ public static void main(String[] args) {
...
+                String newMessage = String.join(" ", wordsWithoutInitiate);
+                String macroName = MacroExecutionModuleImplementation.getMacroName(newMessage, threadId);
+
+                macroCreatorEmail = event.at("/message/sender/email").asText();
+                Map<String, String> macroToCreator = new HashMap<String, String> ();
+                macroToCreator.put(macroName, macroCreatorEmail);
+
+                roomToMacro.put(event.at("/space/name").asText(), macroToCreator);
+                replyText = "The " + macroName + " macro belonging to " + macroCreatorEmail + " has been initiated for this room. All users in this room can use this macro.";
```
lipczak-g: use string format to be consisent.

src/main/java/com/google/flourbot/api/Bot.java
tsa87: I missed the fact that we still need element 0

src/main/java/com/google/flourbot/api/Bot.java
tsa87: Like this one. Oh I see, if there's 2 different messages for DM or a room, should we have 2 constants? 
`"To use a room member's macro, the creator of a macro must write \"@MacroBot initiate MacroName \". " + 
            "Once this has been sent, any room member can use the initialized macro by writing \"@MacroBot MacroName <your message> \" " + 
            "If your macro has already been used in a thread, you may omit the MacroName and can simply write \"@MacroBot <your message>\".";`

src/main/java/com/google/flourbot/api/Bot.java
tsa87: Good idea! If we are refactoring the lines out, would it make sense if the `ExecutionModule` dealt with the whole initialization with `Map<string<Map<string, string>>` and let `Bot.java` only handle getting and sending messages to hangout?

src/main/java/com/google/flourbot/datastorage/FirebaseDataStorage.java
tsa87: Yep just making sure! I like the initialize approach better and I envision that working well with the `Team` structure as well.

SydneyCaulfeild: Ok great! I'll ping you when I've merged this to master. Will you be able to remove the scope input in the website @tsa87?

src/main/java/com/google/flourbot/api/Bot.java
```diff
@@ -28,19 +28,20 @@
...
+
   static final String CHAT_SCOPE = "https://www.googleapis.com/auth/chat.bot";
   private static final String SERVICE_ACCOUNT = "/service-acct.json";
   private static final Logger logger = Logger.getLogger(Bot.class.getName());
-  private String replyText;
-
+  private static final String DM_HELP_MESSAGE = "To use one of your macros, simply write \"@MacroBot MacroName <your message> \".";
+  private static final String ROOM_HELP_MESSAGE = "To use a room member's macro, the creator of a macro must write \"@MacroBot /share MacroName \". " + 
+                                                    "Once this has been sent, any room member can use the initialized macro by writing \"@MacroBot MacroName <your message> \" " + 
+                                                    "If your macro has already been used in a thread, you may omit the MacroName and can simply write \"@MacroBot <your message>\".";
```
tsa87, line 44: nit: \"@MacroBot MacroName your message \" could be replaced with backtick `@MacroBot MacroName <your message>` for better display in hangout

src/main/java/com/google/flourbot/execution/MacroExecutionModule.java
```diff
@@ -4,5 +4,6 @@
 import java.security.GeneralSecurityException;
 
 public interface MacroExecutionModule {
-  String execute(String userEmail, String message, String threadId) throws IOException, GeneralSecurityException;
+  String getReplyText(String[] words, String message, String threadId, String roomId, String messageSenderEmail, String helpMessage) throws IOException, GeneralSecurityException;
```
tsa87: Not clear on the need for both `words` and `message`. Maybe `getRelplyText` can handle `message.split(" ")`?

src/main/java/com/google/flourbot/api/Bot.java
```diff
@@ -57,26 +58,36 @@ public static void main(String[] args) {
...
+          replyText = String.format("Thanks for adding me to %s. Type \"@MacroBot /help\" at any time to see my instructions.", displayName);
+          helpMessage = ROOM_HELP_MESSAGE;
+
         } else {
           String displayName = event.at("/user/displayName").asText();
-          replyText = String.format("Thanks for adding me to a DM, %s!", displayName);
+          replyText = String.format("Thanks for adding me to a DM, %s! Type \"@MacroBot /help\" at any time to see my instructions.", displayName);
+          helpMessage = DM_HELP_MESSAGE;
         }
         break;
```
tsa87: I assumed that we have only one shared Java endpoint to service all the macro instances. Therefore, I don't think we should set the global `helpMessage` variable based on last `AddedToSpace` event, but set the help message reply based on the  `space/type` on `case "Message"`

tsa87 (approved):
Thanks for working on the user scope using the room initialization method! Looks good

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
```diff
@@ -56,23 +55,76 @@ private String getMacroName(String message, String threadId) {
...
+            return "You must type a message when you message me. Please type \"@MacroBot /help\" for more instructions.";
+        }    
+
+        // Reply to the help message.
+        if (words[1].equalsIgnoreCase("/help")) {  
+            return helpMessage;
+        }
+
+        // Get the macro name from the message. 
+        if (words[1].equalsIgnoreCase("/share")) {
```
lipczak-g, line 82: boolean isShareCommand = false;
in the if block: isShareCommand = true;

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
```diff
@@ -56,23 +55,76 @@ private String getMacroName(String message, String threadId) {
...
+        }
+        // If the macro has not been shared to the room, query the database for the macro.
+        else {
+            Optional<Macro> optionalMacro = entityModule.getMacro(messageSenderEmail, macroName);
+            if (!optionalMacro.isPresent()) {
+                return "You do not own/have access to this macro.";
+            }
+            else {
+                // Share the macro if the creator writes "/share".
+                if (words[1].equalsIgnoreCase("/share")) {
```
lipczak-g: if(isShareCommand)

If we have a clumsy method of checking a condition and there is nothing we can do about it, it is better to do it once and keep the state in the variable.

lipczak-g (commented):
the logic looks good now!

src/main/java/com/google/flourbot/api/Bot.java
SydneyCaulfeild: Good idea!

src/main/java/com/google/flourbot/execution/MacroExecutionModule.java
SydneyCaulfeild: I didn't notice this, thanks. I've moved the `message.split` into `getReplyText`.

src/main/java/com/google/flourbot/api/Bot.java
SydneyCaulfeild: Done!

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
SydneyCaulfeild: Sounds good, I've added the variable.

PR #36, opened 2020-07-27
improve card response usage + formatting

miasya:
- add Url to cardResponse.
- remove keyvalue widget. 
- add logic to Bot.java that will return a text message or card depending on what's more appropriate


src/main/java/com/google/flourbot/api/Bot.java
```diff
@@ -79,9 +80,17 @@ public void onEvent(@RequestBody JsonNode event) throws IOException, GeneralSecu
         String threadId = event.at("/message/thread/name").asText();
         
         ChatResponse chatResponse = macroExecutionModule.execute(email, message, threadId);
+        
         replyText = chatResponse.getReplyText();
-        Card card = CardResponse.createCardResponse(replyText);
-        reply.setCards(Collections.singletonList(card));
+        ActionType actionType = chatResponse.getActionType();
+        String documentUrl = chatResponse.getDocumentUrl();
```
tsa87, line 89: What is `documentUrl` here?

tsa87 (approved):
LGTM!

src/main/java/com/google/flourbot/api/Bot.java
miasya: sheetUrl so it creates a button to open up the doc in browser basically

src/main/java/com/google/flourbot/api/Bot.java
```diff
@@ -79,9 +80,17 @@ public void onEvent(@RequestBody JsonNode event) throws IOException, GeneralSecu
...
+
+        // Create card if appropriate
+        if (actionType != null && documentUrl != null) {
```
SydneyCaulfeild, line 92: I like having the logic to only return a card when we need one! I had cards that just returned text for the Resource Helper bot but it looked kind of strange.

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
```diff
@@ -19,6 +19,7 @@
 public class DriveCloudSheet implements CloudSheet {
   private static final String VALUE_INPUT_OPTION = "USER_ENTERED";
   private static final String INSERT_DATA_OPTION = "INSERT_ROWS";
+  private static final String SHEET_NAME = "Sheet1"; // Stretch goal: allow user specified
```
SydneyCaulfeild, line 22: Does this automatically name the sheet they provided in the website to Sheet1?

SydneyCaulfeild (approved):
I really like the card it returns where it shows a screenshot of the current spreadsheet, good work!

src/main/java/com/google/flourbot/api/Bot.java
tsa87: Oh okay neat!

src/main/java/com/google/flourbot/api/Bot.java
lipczak-g: +1 maybe we can have a constant list of action types for which we display cards? Anything that generates some form of structured information, like a list of values from a spreadsheet.

The link to the spreadsheet is a great feature, but I think it should be controlled by the macro creator, sometimes you may not want to disclose to the users that the data is stored in a specific file.

src/main/java/com/google/flourbot/api/DriveCloudSheet.java
miasya: Naw, it just assumes they created the sheet and didn't rename it away from the default

PR #37, opened 2020-07-27
SheetAction inheritance

miasya:
- add abstract SheetAction class that other SheetActions inherit from
- change Action into an interface
- make subclasses SheetAppendRowAction, SheetReadRowAction, SheetReadColumnAction, SheetReadSheetAction
- add handling of subclasses to entitymodule and executionmodule
- remove hardcoded row/col.

next step: add random option for reading row/col/sheet, remove hardcoded sheetname for appending rows, or reading rows and columns

### DIFF
--- a/src/main/java/com/google/flourbot/api/DriveCloudSheet.java
+++ b/src/main/java/com/google/flourbot/api/DriveCloudSheet.java
@@ -78,11 +78,6 @@ public class DriveCloudSheet implements CloudSheet {
     }
   }
 
-  public List<String> readColumn(int column) {
-    // Overload readCol to take a number
-    return readColumn(toAlphabetic(column));
-  }
-
   public List<String> readColumn(String column) {
     List<List<String>> values = readRange(String.format("Sheet1!%s:%s", column, column));
 
--- a/src/main/java/com/google/flourbot/api/DriveCloudSheet.java
+++ b/src/main/java/com/google/flourbot/api/DriveCloudSheet.java
@@ -36,7 +36,7 @@ public class DriveCloudSheet implements CloudSheet {
     return String.format("Sheet1!%s:%s", startColumn, endColumn);
   }
 
-  private String toAlphabetic(int i) {
+  public static String toAlphabetic(int i) {
     // Recursive strategy to build a string that represents the alphabetic name
     // of a column based on column number i
     // e.g. 0 -> A, 1 -> B... 25 -> Z, 26 -> AA...
--- a/src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
+++ b/src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
@@ -110,6 +110,7 @@ public class EntityModuleImplementation implements EntityModule {
             break;
 
           case ("Read Column Action"):
+            // TODO: also handle column as an integer, or just completely remove that logic
             Object column = actionData.get("column");
             if (column == null) {
               throw new IllegalStateException("column not found in Firestore action");
--- a/src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
+++ b/src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
@@ -102,18 +102,27 @@ public class EntityModuleImplementation implements EntityModule {
             break;
  
           case ("Read Row Action"):
-            int row = 1; // TODO: get it from database
-            action = new SheetReadRowAction(sheetUrl, row);
+            Object row = actionData.get("row");
+            if (row == null) {
+              throw new IllegalStateException("row not found in Firestore action");
+            }
+            action = new SheetReadRowAction(sheetUrl, (int) row);
             break;
 
           case ("Read Column Action"):
-            String column = "A"; // TODO: get it from database
-            action = new SheetReadColumnAction(sheetUrl, column);
+            Object column = actionData.get("column");
+            if (column == null) {
+              throw new IllegalStateException("column not found in Firestore action");
+            }
+            action = new SheetReadColumnAction(sheetUrl, (String) column);
             break;
 
           case ("Read Sheet Action"):
-            String sheetName = "Sheet1"; // TODO: get it from database
-            action = new SheetReadSheetAction(sheetUrl, sheetName);
+            Object sheetName = actionData.get("sheetName");
+            if (sheetName == null) {
+              throw new IllegalStateException("sheetName not found in Firestore action");
+            }
+            action = new SheetReadSheetAction(sheetUrl, (String) sheetName);
             break;
 
           case ("Batch Action"):
--- a/src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
+++ b/src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
@@ -4,6 +4,7 @@ import com.google.cloud.firestore.QueryDocumentSnapshot;
 import com.google.flourbot.datastorage.DataStorage;
 import com.google.flourbot.entity.action.Action;
 import com.google.flourbot.entity.action.sheet.SheetAppendRowAction;
+import com.google.flourbot.entity.action.sheet.SheetReadColumnAction;
 import com.google.flourbot.entity.action.sheet.SheetReadRowAction;
 import com.google.flourbot.entity.action.sheet.SheetEntryType;
 import com.google.flourbot.entity.trigger.CommandTrigger;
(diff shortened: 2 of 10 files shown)


src/main/java/com/google/flourbot/api/DriveCloudSheet.java
```diff
@@ -36,7 +36,7 @@ private String getRange(int numEntryColumns, int numSkipColumns) {
     return String.format("Sheet1!%s:%s", startColumn, endColumn);
   }
 
-  private String toAlphabetic(int i) {
+  public static String toAlphabetic(int i) {
```
SydneyCaulfeild, line 39: Like the recursion!

src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
```diff
@@ -96,15 +98,27 @@ public EntityModuleImplementation(DataStorage dataStorage) {
 
         switch(sheetAction) {
           case ("Sheet Append Action"):
-            action = new SheetAppendRowAction(columnValue, sheetAction, sheetUrl);
+            action = new SheetAppendRowAction(sheetUrl, columnValue);
             break;
  
           case ("Read Row Action"):
-            action = new SheetReadRowAction(columnValue, sheetAction, sheetUrl);
+            int row = 1; // TODO: get it from database
```
SydneyCaulfeild: Are you planning on making a new class for these queries related to spreadsheet appends/reads, or just make individual functions in this class for the queries?

src/main/java/com/google/flourbot/entity/action/Action.java
```diff
@@ -1,7 +1,7 @@
 package com.google.flourbot.entity.action;
 
-public abstract class Action {
-  public abstract ActionType getActionType();
-  public abstract String getSheetId();
-  public abstract String getSheetUrl();
+public interface Action {
```
SydneyCaulfeild: Just out of curiosity, why did we switch it from abstract class to an interface?

tsa87 (approved):
Inheritance Looks great!

src/main/java/com/google/flourbot/entity/action/sheet/SheetAction.java
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  public abstract ActionType getActionType();
+
+  public String getSheetUrl() {
+    return sheetUrl;
+  }
+
+  public String getSheetId() {
+    String sheetId = "";
```
SydneyCaulfeild: Is there a way we can throw an exception instead of returning an empty sheedId here?

src/main/java/com/google/flourbot/entity/action/sheet/SheetReadColumnAction.java
```diff
@@ -0,0 +1,29 @@
...
+
+  private final ActionType actionType = ActionType.SHEET_READ_COLUMN;
+  private final String column;
+
+  public SheetReadColumnAction(String sheetUrl, String column) {
+    super(sheetUrl);
+    this.column = column;
+  }
+
+  public SheetReadColumnAction(String sheetUrl, int column) {
```
SydneyCaulfeild, line 16: When would we use a string vs an int?

src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
miasya: Each action is a different subclass of SheetAction, and so the subclasses add in methods and attributes that are more specific to what they handle

src/main/java/com/google/flourbot/entity/action/Action.java
miasya: Good question! I want to see what other people say, but my reasoning was:

- Interfaces cannot have implementation of the methods, so we would want SheetAction to be an abstract class for sure. This doesn't tell us anything about Action yet.

- Since SheetAction is an abstract class, and SheetAppendRowAction extends it, SheetAppendRowAction therefore inherits the (non-abstract) methods that were implemented in SheetAction, and has to implement the abstract methods of SheetAction.
 - Also, in an abstract class, we can have private and protected members, which we can't have in an interface.

- Action doesn't really have anything (no functionality) that we want SheetAction to inherit, instead, it provides the method signatures that we need to implement and nothing else. Action simply provides a common structure that classes need to follow. Which to make makes more sense to have as an interface

src/main/java/com/google/flourbot/entity/action/sheet/SheetAction.java
miasya: I like that you're thinking about ways to make this more robust. getSheetId() will always work if the URL is properly supplied. So maybe it would be more appropriate to do URL checking when the user first makes their bot.

Alternatively, we could catch a bad URL later on when we handle the case where the bot doesn't have permission to write to the desired document. -- I think we're doing this soon

src/main/java/com/google/flourbot/entity/action/sheet/SheetReadColumnAction.java
miasya: This was written to make sure it accepts both formats, but actually yeah, the int part isn't as important if we can make sure a String is what we have in Firestore. Currently, other code in entitymoduleimplementation assumes that it's always a string, but soon I hope we can add an option in the UI where a user can supply either one.

src/main/java/com/google/flourbot/entity/action/Action.java
SydneyCaulfeild: Ok, thanks! And ya that's a good idea to use protected for things like the sheet url.

src/main/java/com/google/flourbot/entity/action/sheet/SheetAction.java
SydneyCaulfeild: Ok sounds good, I didn't now the getSheetId always works with a proper URL

src/main/java/com/google/flourbot/entity/action/sheet/SheetReadColumnAction.java
SydneyCaulfeild: Ok! I like giving users the options, my only concern is will this mean we will have null values in the database? Will this be a problem?

src/main/java/com/google/flourbot/entity/action/Action.java
lipczak-g: Interesting discussion. In my mind, Interface is more applicable when we have multiple implementations that have the same functionality but fulfilled in a different way. Like implementation that talks to Drive and Office365 to operate on a spreadsheet. Abstract class is more applicable when you have a hierarchy of object types in which each level of the hierarchy is more specific than the parent. So here abstract class makes more sense to me here...

src/main/java/com/google/flourbot/entity/action/Action.java
```diff
@@ -1,7 +1,7 @@
...
+  public ActionType getActionType();
+  public String getSheetId();
```
lipczak-g: no matter if this is an interface or abstract class, the sheet specific methods don't belong at the Action level. What if we add DocAction?

src/main/java/com/google/flourbot/entity/action/sheet/SheetAction.java
```diff
@@ -0,0 +1,33 @@
...
+    Pattern pattern = Pattern.compile("(d\\/[a-zA-Z0-9-_]+)");
```
lipczak-g: make the pattern static final at the class level.

src/main/java/com/google/flourbot/entity/action/sheet/SheetAction.java
```diff
@@ -0,0 +1,33 @@
...
+    Matcher matcher = pattern.matcher(this.sheetUrl);
+    // Take the first occurance
```
lipczak-g: nit: occurence

src/main/java/com/google/flourbot/entity/action/sheet/SheetReadColumnAction.java
```diff
@@ -0,0 +1,29 @@
...
+    // Overload
+    super(sheetUrl);
+    this.column = DriveCloudSheet.toAlphabetic(column + 1);
```
lipczak-g, line 19: not clear why we have +1 here.

src/main/java/com/google/flourbot/entity/action/Action.java
miasya: Changed it!

src/main/java/com/google/flourbot/entity/action/sheet/SheetReadColumnAction.java
miasya: Oh if they key isn't there, it returns null, and we can handle that! So we won't have nulls in the firestore

src/main/java/com/google/flourbot/entity/action/Action.java
miasya: renamed!

src/main/java/com/google/flourbot/entity/action/sheet/SheetReadColumnAction.java
miasya: The function is A-> 0 but here we want A->1

PR #39, opened 2020-07-27
Enable Team Structure in UI

tsa87:
- Created `Team` class in Dart that represents a public chatbot with multiple functionalities (macros).
- Changed the Flutter UI to allow the action to create a new team object and upload the result.
- Created the associated state management for team form.
- Added options for `Map Action` and `URL Redirect Action`
- Refactor Firestore functions into a Database module.


flutter_src/lib/logic/api/firestore_db.dart
```diff
@@ -0,0 +1,59 @@
+import 'package:cloud_firestore/cloud_firestore.dart';
+import 'package:macrobaseapp/logic/state/macro_state/macro_notifier.dart';
+import 'package:macrobaseapp/model/adapters/macro_model.dart';
+import 'package:macrobaseapp/model/entities/macro.dart';
+
+class FirestoreService {
```
lipczak-g, line 6: Uhhh... not sure if there is an option to do in GitHub, but it would be much easier to understand if you changed the original file with this logic and not delete old/add new.

flutter_src/lib/logic/state/bloc_validator.dart
```diff
@@ -1,20 +1,37 @@
...
-    List<Macro> list = await queryMacro(name);
+    List<Macro> list = await db.queryMacro(name);
     if (list.length > 0) {
       return "Sorry, this macro name already exist";
     }
     return null;
   }
 
+  static String oneWordValidator(String string) {
+    if (string.split(" ").length > 1) {
```
lipczak-g: I'm guessing there must be an easier way like "contains" method.

flutter_src/lib/logic/state/bloc_validator.dart
```diff
@@ -1,20 +1,37 @@
...
+      return "Macro Name cannot have space in between.";
+    }
+    return null;
+  }
+
+  static String UrlValidator(String string) {
+    bool isValidUrl = Uri.parse(string).isAbsolute;
+    if (!isValidUrl) {
+      return string + " is not a valid URL to an ";
```
lipczak-g: ...to an ???

flutter_src/lib/logic/state/bloc_validator.dart
```diff
@@ -1,20 +1,37 @@
...
 class CustomBlocValidator {
   static Future<String> nameValidator(String name) async {
+    FirestoreService db = FirestoreService();
+
     //Avoid too many Firestore calls
     await Future.delayed(Duration(milliseconds: 500));
-    List<Macro> list = await queryMacro(name);
+    List<Macro> list = await db.queryMacro(name);
     if (list.length > 0) {
       return "Sorry, this macro name already exist";
```
lipczak-g: Sorry for missing this before, but overall this is not a good validator structure. The validator should only return true/false. This way it can be reused in other tasks. Defining the specific string that is displayed is the job of the caller. This applies to all the validators here.

flutter_src/lib/logic/state/bloc_validator.dart
```diff
@@ -1,20 +1,37 @@
...
   }
 
+  static String oneWordValidator(String string) {
+    if (string.split(" ").length > 1) {
+      return "Macro Name cannot have space in between.";
+    }
+    return null;
+  }
+
+  static String UrlValidator(String string) {
```
lipczak-g: nit: lower case u to start the method name.

flutter_src/lib/logic/state/bloc_validator.dart
```diff
@@ -31,12 +48,14 @@ class CustomBlocValidator {
     List<String> emails = string.split(',');
 
     for (int i = 0; i < emails.length; i++) {
-      bool emailValid = RegExp(r"^[a-zA-Z0-9.a-zA-Z0-9.!#$%&'*+-/=?^_`{|}~]+@[a-zA-Z0-9]+\.[a-zA-Z]+").hasMatch(emails[i]);
+      bool emailValid = RegExp(
```
lipczak-g, line 51: nit: typically you want to trim whitespaces from emails[i] here in case the user added spaces after the comma.

flutter_src/lib/logic/state/bloc_validator.dart
```diff
@@ -31,12 +48,14 @@ class CustomBlocValidator {
...
+              r"^[a-zA-Z0-9.a-zA-Z0-9.!#$%&'*+-/=?^_`{|}~]+@[a-zA-Z0-9]+\.[a-zA-Z]+")
+          .hasMatch(emails[i]);
       if (!emailValid) {
         errors.add(emails[i] + " is not an valid email.");
```
lipczak-g: no need to be so specific, still better to return just false and have a generic failure message.

lipczak-g (commented):
wow, amazing work, sorry I was able to look carefully only at a couple of files, the night is too short :P

flutter_src/lib/logic/state/bloc_validator.dart
Vector-S: Agree to simplify validator output to boolean. Consider adding a logger to log details.

Vector-S (approved):

PR #40, opened 2020-07-28
Add random option to read row, column, sheet

miasya:
- Now the Firestore will look for a boolean called selectRandomEntry under action for the read row, column, and sheet actions


tsa87 (approved):
LGTM!

src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
```diff
@@ -106,7 +106,7 @@ public EntityModuleImplementation(DataStorage dataStorage) {
             if (row == null) {
               throw new IllegalStateException("row not found in Firestore action");
             }
-            action = new SheetReadRowAction(sheetUrl, (int) row);
+            action = new SheetReadRowAction(sheetUrl, (int) row, (boolean) actionData.get("selectRandomEntry"));
```
lipczak-g: This works for now, but it doesn't scale, I think it would be better to have a SelectionMethod enum that contains other options, like first, last etc

lipczak-g: Also, please see my comment below, but I think this approach is not intuitive. The action is called ReadRow, but depending on whether selectRandomEntry is false/true you will either return the full row or just one field.

I think it would be better to have the enum as described above and an int variable called limit. -1 would mean no limit. This way you can define results like 
FIRST -1 (all fields in the natural order - the default)
LAST 3 - the last 3 fields reverted from the end
RANDOM 1 - single random field (which is what you have here)
This way you preserve the nature of the action (i.e. get the list of fields in the row) but also introduce some flexibility.
For now you can just hardcode two options: FIRST -1 and RANDOM 1 and pick them depending on the state of selectRandomEntry - later we can fix the UI.

src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
```diff
@@ -115,15 +115,15 @@ public EntityModuleImplementation(DataStorage dataStorage) {
...
+            action = new SheetReadColumnAction(sheetUrl, (String) column, (boolean) actionData.get("selectRandomEntry"));
             break;
 
           case ("Read Sheet Action"):
             Object sheetName = actionData.get("sheetName");
             if (sheetName == null) {
               throw new IllegalStateException("sheetName not found in Firestore action");
             }
-            action = new SheetReadSheetAction(sheetUrl, (String) sheetName);
+            action = new SheetReadSheetAction(sheetUrl, (String) sheetName, (boolean) actionData.get("selectRandomEntry"));
```
lipczak-g: I don't see how random entry applies to read sheet, can you clarify?

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
```diff
@@ -146,19 +146,33 @@ public ChatResponse execute(String userEmail, String message, String threadId) t
       case SHEET_READ_ROW:
         int row = ((SheetReadRowAction) action).getRow();
```
lipczak-g, line 203: getRow should have all of this special selection logic hidden in it. So you call getRow and you always get a list, even if this happens to be one random field from the list.

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
```diff
@@ -146,19 +146,33 @@ public ChatResponse execute(String userEmail, String message, String threadId) t
...
+        }
         break;
 
       case SHEET_READ_SHEET:
         String sheetName = ((SheetReadSheetAction) action).getSheetName();
         List<List<String>> sheetData = cloudSheet.readSheet(sheetName);
-        chatResponse = ChatResponse.createChatResponseWithListList(sheetData, actionType, documentUrl);
+
+        if (((SheetReadSheetAction) action).getSelectRandomEntry()) {
+          chatResponse = new ChatResponse(selectRandomFromListList(sheetData), actionType, documentUrl);
```
lipczak-g: uhhh... I can't think of a use case for this... looks complex to implement as well. How about we skip it for now?

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
```diff
@@ -146,19 +146,33 @@ public ChatResponse execute(String userEmail, String message, String threadId) t
       case SHEET_READ_ROW:
         int row = ((SheetReadRowAction) action).getRow();
         List<String> rowData = cloudSheet.readRow(row);
-        chatResponse = ChatResponse.createChatResponseWithList(rowData, actionType, documentUrl);
+        if (((SheetReadRowAction) action).getSelectRandomEntry()) {
+          chatResponse = new ChatResponse(selectRandomFromList(rowData), actionType, documentUrl);
```
lipczak-g: If you need to call a different constructor if the list has a single element use rowData.length() == 1 as the condition instead.

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
```diff
@@ -177,8 +191,14 @@ private String getDate(String pattern) {
     return myDateObj.format(myFormatObj);
   }
 
-  private String selectRandomFromData(List<String> data) {
+  private String selectRandomFromList(List<String> data) {
```
lipczak-g: this method doesn't really belong here. This a good reason to create SheetAction class and add protected methods that will execute this FIRST/LAST/RANDOM N logic.

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
```diff
@@ -177,8 +191,14 @@ private String getDate(String pattern) {
...
     Random rand = new Random();
     return data.get(rand.nextInt(data.size()));
   }
+
+  private String selectRandomFromListList(List<List<String>> data) {
```
SydneyCaulfeild: Why are we using a nested list here? What is the outer list holding?

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
miasya: removed!

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
miasya: It was originally to select a random entry from a sheet, which the big list holds little lists (and the little lists are the rows)

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
miasya: Great idea! Moving it

src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
miasya: I love the enum idea! Added it. Will need to code the options later - if we judge it is important
Note that I moved the handling logic to DriveCloudSheet since I thought it was appropriate to have there, but if you think it's better to have within the SheetReadRowAction and SheetReadColumnAction let me know

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
miasya: I changed it to return a list with a single entry, so the same constructor can be used

PR #41, opened 2020-07-29
Team Assistant Feature

tsa87:
WIP

- Display nested macros in Flutter
- Create new notifier and util functions in Dart
- Upcoming Server additions...

### DIFF
--- a/flutter_src/lib/logic/api/firestore_db.dart
+++ b/flutter_src/lib/logic/api/firestore_db.dart
@@ -24,13 +24,9 @@ class FirestoreService {
     List<Macro> _macroList = [];
 
     snapshot.documents.forEach((element) {
-      try {
         Macro macro = MacroModel.fromJson(element.data);
         macro.macroId = element.documentID;
         _macroList.add(macro);
-      } catch (e) {
-        print(element.data);
-      }
     });
 
     macroNotifier.macroList = _macroList;
@@ -42,13 +38,9 @@ class FirestoreService {
     List<Team> _teamList = [];
 
     snapshot.documents.forEach((element) {
-      try{
         Team team = TeamModel.fromJson(element.data);
         team.teamId = element.documentID;
         _teamList.add(team);
-      } catch(e) {
-        print(element.data);
-      }
     });
 
     teamNotifier.teamList = _teamList;
--- a/flutter_src/lib/logic/api/firestore_db.dart
+++ b/flutter_src/lib/logic/api/firestore_db.dart
@@ -9,38 +9,43 @@ import 'package:macrobaseapp/model/entities/team.dart';
 class FirestoreService {
   final Firestore db;
 
-  FirestoreService({FirestoreService db})
-    : db = db ?? Firestore.instance;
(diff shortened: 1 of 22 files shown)


flutter_src/lib/logic/api/firestore_db.dart
```diff
@@ -1,32 +1,58 @@
...
+        Team team = TeamModel.fromJson(element.data);
+        team.teamId = element.documentID;
+        _teamList.add(team);
+      } catch(e) {
+        print(element.data);
+      }
+    });
+
+    teamNotifier.teamList = _teamList;
+  }
```
tsa87: These two functions look really similar, however, I wasn't sure how to combine them into one generic function.

flutter_src/lib/logic/state/team_state/team_notifier.dart
```diff
@@ -0,0 +1,34 @@
+import 'dart:collection';
+import 'package:flutter/material.dart';
+import 'package:macrobaseapp/model/entities/team.dart';
+
+class TeamNotifier with ChangeNotifier {
```
tsa87, line 5: Also too similar to `macro_notifier.dart`

flutter_src/lib/logic/api/firestore_db.dart
lipczak-g: yeah, I can see the structural similarity, but the types make them fairly different, so I think they are better to be kept separated. BTW, you fixed the style in the top onem but not the bottom one.

lipczak-g: Just FYI: In Java you would use generics to do it, but again, not worth the effort here... too simple
https://docs.oracle.com/javase/tutorial/extra/generics/methods.html

flutter_src/lib/logic/state/team_state/team_notifier.dart
lipczak-g: Again generic class could be a solution here, but it is not worth here. Sometimes clarity is more important than concise code.

flutter_src/lib/model/entities/action.dart
```diff
@@ -67,7 +67,11 @@ class BatchAction extends SheetAction {
 }
 
 class AppendAction extends SheetAction {
-  static const List<String> VALUE_LIST = ["EMPTY", "TIME", "EMAIL", "CONTENT"];
+  static const List<String> VALUE_LIST = [VALUE_EMPTY, VALUE_TIME, VALUE_EMAIL, VALUE_CONTENT];
+  static const VALUE_EMPTY = "EMPTY";
```
lipczak-g, line 71: wonder how this works when VALUE_* consts are defined after VALUE_LIST

flutter_src/lib/logic/api/firestore_db.dart
tsa87: Thanks for letting me know about Generics! Yea, in this case it seems okay to keep it this way for now.

flutter_src/lib/model/entities/action.dart
tsa87: Changed the order

PR #42, opened 2020-07-29
Remove the card subtitle and word TODO from card display.

SydneyCaulfeild:
Also fix the message "The reply text we got is" on card display.


tsa87 (approved):
LGTM!

PR #44, opened 2020-07-31
Add tests for DriveCloudSheet.java

### LINKED ISSUE #32: improve exception handling in CloudDocClient
Use try catch

miasya:


src/test/java/com/google/flourbot/api/DriveCloudSheetTest.java
```diff
@@ -0,0 +1,94 @@
...
+  
+  private CloudDocClient cloudDocClient;
+  private CloudSheet cloudSheet;
+
+  @Before
+  public void setUp() {
+    cloudDocClient = new DriveClient();
+    try {
+      cloudSheet = cloudDocClient.getCloudSheet(SPREADSHEET_ID);
+    } catch (IOException e) {
```
lipczak-g, line 76: I don't understand what is the reason for this catches. In general, your test should be stable enough, so this never happens. Also, setup should not have any assertions - the fail will be communicated with the exception anyway.

src/test/java/com/google/flourbot/api/DriveCloudSheetTest.java
```diff
@@ -0,0 +1,94 @@
...
+  private static final String RANGE_ROW = "2";
+  private static final String RANGE_COLUMN = "C";
+  //TODO: add invalid range options
+  
+  private CloudDocClient cloudDocClient;
+  private CloudSheet cloudSheet;
+
+  @Before
+  public void setUp() {
+    cloudDocClient = new DriveClient();
```
lipczak-g: Do I get it right that you are establishing full connection to Drive API here? This is a potential source of failures unrelated to the tested code. Typically in a unit test you want to eliminate such dependencies.

src/test/java/com/google/flourbot/api/DriveCloudSheetTest.java
```diff
@@ -0,0 +1,94 @@
...
+    try {
+      cloudSheet = cloudDocClient.getCloudSheet(SPREADSHEET_ID);
```
lipczak-g: Hmmm... I don't get how are you able to get the cloudSheet object with empty id.

I think instead you should call new DriveCloudSheet(String documentId, Sheets documentService) with the mock of Sheets object. This way you can take DriveClient out of the picture.

lipczak-g (commented):
Couple quick comments to unblock you for today.

src/test/java/com/google/flourbot/api/DriveCloudSheetTest.java
miasya: Updated to handle exceptions in try catch in cloud doc client - so I think that should throw the proper exceptions and cause the fail without the try catch nor assertions in the setup?

src/test/java/com/google/flourbot/api/DriveCloudSheetTest.java
miasya: Yes, but I think you're completely right! Will update

src/test/java/com/google/flourbot/api/DriveCloudSheetTest.java
miasya: added mocks!

src/test/java/com/google/flourbot/api/DriveCloudSheetTest.java
```diff
@@ -0,0 +1,327 @@
...
+  @Test
+  public void testReadRangeSingle() {
+    String[] array = {"B1"};
+    List<List<Object>> values = Arrays.asList(
+      Arrays.asList(array)
+    );
+
+    when(sheetsServiceMock.spreadsheets()).thenReturn(spreadsheetsMock);
+    when(spreadsheetsMock.values()).thenReturn(valuesMock);
+    try {
```
lipczak-g, line 73: don't understand why the try is needed, but if it is needed please add Assert.fail in the catch block.

src/test/java/com/google/flourbot/api/DriveCloudSheetTest.java
```diff
@@ -0,0 +1,327 @@
...
+      when(valuesMock.get(SPREADSHEET_ID, String.format("%s!%s", SHEET_NAME, RANGE_SINGLE))).thenReturn(getMock);
+      when(getMock.execute()).thenReturn(valueRangeMock);
+    } catch (IOException e) {}
+
+    when(valueRangeMock.getValues()).thenReturn(values);
```
lipczak-g, line 80: super nit: I think you should be able to create a setUpSpreadsheets method where some or all of these could be added, but this is not a hard requirement and some people argue that the setup should be explicit in all test functions even if it creates code bloat.

lipczak-g (approved):

PR #45, opened 2020-08-01
Flutter Testing and Miscellaneous

tsa87:


flutter_src/lib/presentation/pages/login_page.dart
```diff
@@ -23,11 +23,11 @@ class LoginPageBody extends StatelessWidget {
             mainAxisAlignment: MainAxisAlignment.center,
             children: <Widget>[
               Text(
-                "Macrobase",
+                "Macrobot",
                 style: Theme.of(context).textTheme.headline4,
               ),
               Text(
-                "Automate the boring stuff.",
```
lipczak-g, line 30: I liked this one ;)

flutter_src/lib/presentation/navigation/main_navigator.dart
```diff
@@ -35,11 +35,11 @@ class _MainNavigatorState extends State<MainNavigator> {
             tabs: [
               Tab(
                 icon: Icon(Icons.add),
-                text: 'New Macro',
+                text: 'New Chatbots',
```
lipczak-g: We have to be really careful with what we promise. Using Chatbot as the keyword can give users the impression that we allow them to create full bots which is not the case.

tsa87: True, I changed it back to bots.

PR #47, opened 2020-08-04
Tests

SydneyCaulfeild:


src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
```diff
@@ -122,13 +126,20 @@ public void removeMacro(String roomId) {
       roomToMacro.remove(roomId);
   }
 
+  // Used for testing
```
lipczak-g, line 129: In general we should avoid using methods that are specifically for testing... but it is OK to keep.

src/test/java/MacroExecutionModuleImplementationTest.java
```diff
@@ -0,0 +1,154 @@
...
+    private static final String HELP_MESSAGE = "This is your help message";
+
+    // TODO: use global vars to initialize DEFAULT_MACRO 
+    SheetEntryType[] columnHeaders = {SheetEntryType.TIME, SheetEntryType.EMAIL, SheetEntryType.CONTENT};
+    private static final Macro DEFAULT_MACRO = new Macro("scaulfeild@google.com", "DailyBot", new CommandTrigger("Command Trigger"), new SheetReadSheetAction("SheetUrl", "SheetName")); 
+
+    private static MacroExecutionModuleImplementation execution;
+    private static EntityModule mockEntityModule;
+    private static CloudDocClient mockCloudDocClient;
+    private static MacroExecutionModuleImplementation mockMacroExecutionModuleImplementation;
```
lipczak-g: you don't need this one.

src/test/java/MacroExecutionModuleImplementationTest.java
```diff
@@ -0,0 +1,154 @@
...
+        execution  = MacroExecutionModuleImplementation.initializeServer(mockEntityModule, mockCloudDocClient);
+    }
+    
+    // Remove a macro from the roomToMacro Map
+    @Test
+    public void testRemoveMacro() {
+        Map<String, Map<String, String>> actual = execution.getRoomToMacro();
+        Map<String, String> macro = new HashMap<String, String>();
+        macro.put(MACRO_NAME, USER_EMAIL);
+        actual.put(ROOM1, macro);
```
lipczak-g: This should be done by the method that really adds the room, this can be a helper method if you are going to use it multiple times. But again, it is OK to leave as is... don't want you to spend too much time on it.

src/test/java/MacroExecutionModuleImplementationTest.java
```diff
@@ -0,0 +1,154 @@
...
+        ChatResponse chatResponse = execution.getReplyText("@MacroBot", THREAD1, ROOM1, USER_EMAIL, HELP_MESSAGE);
+        String actual = chatResponse.getReplyText();
+        Assert.assertEquals(EMPTY_MESSAGE_RESPONSE, actual);
+    }
+
+    // Test using a macro for the first time when you are the creator. In this case, scaulfeild@google.com created DailyBot
+    @Test
+    public void useYourMacroInNewRoom() throws IOException, GeneralSecurityException, Exception {
+        
+        //when(mockEntityModule.getMacro("scaulfeild@google.com", "DailyBot")).thenReturn(Optional.of(DEFAULT_MACRO));
```
lipczak-g: not clear how does it work without mocking

src/test/java/MacroExecutionModuleImplementationTest.java
```diff
@@ -0,0 +1,154 @@
...
+    
+        ChatResponse chatResponse = execution.getReplyText("@MacroBot DailyBot please log my message", THREAD1, ROOM1, "scaulfeild@google.com", HELP_MESSAGE);
+        String actual = chatResponse.getReplyText();
+        Assert.assertEquals("successfully executed", actual);
```
lipczak-g: This should be a constant.

src/test/java/MacroExecutionModuleImplementationTest.java
```diff
@@ -0,0 +1,154 @@
...
+    public void usingAPrivateMacro() throws IOException, GeneralSecurityException {
+        ChatResponse chatResponse = execution.getReplyText("@MacroBot CongraBot hi", "spaces/AAAAAAAAAAA/threads/BBBBBBBBBBB", "spaces/AAAAAAAAAAA", "scaulfeild@google.com", HELP_MESSAGE);
+        String actual = chatResponse.getReplyText();
+        String macroName = execution.getMacroName("@MacroBot CongraBot hi", "spaces/AAAAAAAAAAA/threads/BBBBBBBBBBB");
+        String expected = (new ChatResponse(String.format("You do not own/have access to %s.", macroName))).getReplyText();
+
+        Assert.assertEquals(expected, actual);
+    }*/
+
+    // Test using someone else's macro after it has been shared
```
lipczak-g: Yeah that's the pattern that we should follow instead of testing the maps directly.

src/test/java/MacroExecutionModuleImplementationTest.java
```diff
@@ -0,0 +1,187 @@
...
+    // Test using a macro for the first time when you are the creator. Testing with SheetAppendRow as action type. 
+    @Test
+    public void useYourMacroInNewRoom() throws IOException, GeneralSecurityException, Exception {
+        List<String> values = new ArrayList<String>();
+        values.add(getDate("dd-MM-yyyy HH:mm:ss"));
+        values.add(USER_EMAIL);
+        //TODO: change this message to use a var instead of hardcoded
+        values.add("please log my message");
+        
+        when(mockEntityModule.getMacro(USER_EMAIL, MACRO_NAME)).thenReturn(Optional.of(DEFAULT_MACRO));
```
lipczak-g: The null pointer exception comes from the fact that DEFAULT_MACRO is disconnected from mockAction. To make it work you would have to somehow add the mockAction object to the DEFAULT_MACRO object. But there is an easier and better approach. Simply initialize DEFAULT_MACRO with proper sheetUrl. As you can see the documentId comes from the documentUrl [1]. If I understand the regex correctly making sheetUrl "d/sheetId" should be enough to make things working.

[1] https://github.com/googleinterns/step81-2020/blob/master/src/main/java/com/google/flourbot/entity/action/sheet/SheetAction.java

src/main/java/com/google/flourbot/execution/MacroExecutionModuleImplementation.java
```diff
@@ -222,10 +217,28 @@ private ChatResponse execute(String userEmail, String macroCreatorEmail, String
         throw new IllegalStateException(
             "Action type named: " + actionType.toString() + "is not implemented yet!");
     }
-
     return chatResponse;
   }
 
+  public static String getMessageContent(String message, String macroName) {
```
lipczak-g, line 223: it is fine to extract this logic, but it shouldn't be public

src/test/java/MacroExecutionModuleImplementationTest.java
```diff
@@ -0,0 +1,236 @@
...
+
+    private String createDefaultChatResponse(String userEmail) {
+        List<String> values = new ArrayList<String>();
+        values.add(getDate("dd-MM-yyyy HH:mm:ss"));
+        values.add(userEmail);
+        values.add(MESSAGE_CONTENT_ONLY);
+        return (ChatResponse.createChatResponseWithList(values, DEFAULT_MACRO.getAction().getActionType(), DEFAULT_MACRO.getAction().getDocumentUrl())).getReplyText();
+    }
+
+    private void setUpMocks() throws IOException, GeneralSecurityException {
```
lipczak-g, line 230: super nit: the name of this method should be more descriptive - what exactly are you setting up?

src/test/java/MacroExecutionModuleImplementationTest.java
```diff
@@ -0,0 +1,236 @@
...
+        execution.getThreadMacroMap().clear();
+        execution.getRoomToMacro().clear();
+    }
+    
+    // Remove a macro from the roomToMacro Map.
+    @Test
+    public void testRemoveMacro() {
+        shareDefaultMacroToRoom(ROOM1);
+
+        execution.removeMacro(ROOM1);
```
lipczak-g, line 89: nit: so here is an example of a place in which it would be good to assert the macro got indeed stored. Otherwise the final assert could pass even without calling execute.

lipczak-g (approved):
Nice! I really like how the test file turned out. It is clean and easy to follow which is important for tests.

PR #48, opened 2020-08-04
Mock modules

tsa87:
Added Mock Objects (StubDataStorage, StubDriveClient and StubSheet) that can be used for unit testing - decoupling dependency and avoiding external API calls

### DIFF
--- a/src/main/java/com/google/flourbot/api/StubClient.java
+++ b/src/main/java/com/google/flourbot/api/StubClient.java
@@ -11,6 +11,4 @@ public class StubClient implements CloudDocClient{
   public CloudSheet getCloudSheet(String documentId){
     return this.stubSheet;
   }
-
-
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/flourbot/api/StubClient.java
```diff
@@ -0,0 +1,14 @@
+package com.google.flourbot.api;
+
+public class StubClient implements CloudDocClient{
```
lipczak-g, line 3: Interesting, what is the advantage of stubs over Mockito's mocks?

tsa87: It was taught in one of the boarding training workshops for STEP. I guess the only advantage is learning how Mockito is maybe implemented behind the scene.

PR #49, opened 2020-08-04
Entity module test

tsa87:


src/test/java/com/google/flourbot/entity/EntityModuleImplementationTest.java
```diff
@@ -0,0 +1,104 @@
...
+import java.io.IOException;
+import java.net.URL;
+import java.util.Map;
+import java.util.Optional;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class EntityModuleImplementationTest {
+
+  private StubDataStorage dataStorage;
```
lipczak-g, line 21: it looks to me as a good use case for Mockito you can mock the data storage interface and then use when().thenReturn() pattern. Please check the other test PRs for some examples.

src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
```diff
@@ -55,80 +56,86 @@ public EntityModuleImplementation(DataStorage dataStorage) {
...
-        for (String type : columnStringList) {
+    try {
+      String actionType = (String) actionData.get("type");
+
+      switch (actionType) {
+        case ("Sheet Action"):
+          ArrayList<String> columnStringList = (ArrayList<String>) actionData.get("columnValue");
+          // Converts into ENUM type
+          ArrayList<SheetEntryType> columnTypeList = new ArrayList<>();
+          for (String type : columnStringList) {
```
lipczak-g, line 86: nit: I assume this is just formatting clean-up - these should be submitted separately to avoid confusion.

lipczak-g (commented):
Overall the tests look good, but IMO we should use Mockito instead of self made stubs.

src/main/java/com/google/flourbot/entity/EntityModuleImplementation.java
tsa87: I was adding the indents for a try catch for NullPointerException. Hmm, It seems like git recognized it as adding and deleting new lines

tsa87: I will also keep in mind to reformat in a separate branch in the future!

src/test/java/com/google/flourbot/entity/EntityModuleImplementationTest.java
tsa87: Sure, I can look into replacing Stubs with Mockito!

